PR #1, opened 2017-05-12
started encoding of Eusebius letter

PietroLiuzzo:
Not sure how to deal with the text commented out and if the numbered sections are actually lines or should be sections. The sigla correspond to what? and the | signs?


Gnisci:
I think you need to treat letter and canons as separate entities (because they don't always come together), you treat the letter as one text, and the tables as a text with a particular layout (it is tabular in form, with sequences of numbers) or as a diagram. Finally, you add the keyword "arch" to the "art keywords" and you encode (when such instances occur) that arches decorate both of either

PietroLiuzzo:
@Gnisci this was a very old test of a pull request which I did with a text provided by Alessandro.
I have pushed to do some cleaning of things left half way. please feel to edit and correct the entry in the usual ways.

PR #2, opened 2017-10-30
Correction from ጽድቅ፡ (righteousness) to ድድቅ፡ (misfortune)

smaugustine:
This would be one of the typos HaCohen mentions as likely being present in his text of the Psalms, based on Ludolf's edition.


abausi:
He is absoluitely right: it must be corrected into ድድቅ፡, under which entry (by the way) our good Dillmann has the right reference to Psalm 90:6 and also gives the Greek equivalent σύμπτωμα, that is the text of the LXX version.

PR #3, opened 2018-02-25
An OCR error: ኬትከ፡ for ቤትከ፡ (Thy house)

smaugustine:
Today's məsbak, no less (Ps. 68:12).


abausi:
Correct, it must be ቤትከ፡

PR #5, opened 2019-11-04
Ezekiel etc 

thea-m:
created recensions of Ezekiel according to Knibb, created records for the Lives of the Prophets


thea-m:
and deleted double records from chaines rep.

PR #10, opened 2019-11-19
added incipit version to LIT2730RepCh14

eu-genia:


2001-3000/LIT2730RepCh14.xml
```diff
@@ -85,6 +84,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                </ab>
             </div>
          </div>
+         <div type="edition" xml:lang="gez">
+            <note>The following incipit is according to  <ref type="mss" corresp="ESath007#a4"></ref>
+            </note>
+            <div type="textpart" subtype="incipit">
+               <ab>ሰላም፡ ለኪ፡ ማርያም፡ ስጣሕት፡ ቀስተተ፡ መሐላሁ፡ ለኖኅ፡…</ab>
```
thea-m: I would delete the ..., since it is marked up as an incipit, it should be obvious that more text follows?

eu-genia: done

PR #11, opened 2019-11-19
Minor edits to authorship + New record

eu-genia:


new/LIT5810HomilySatan.xml
```diff
@@ -0,0 +1,65 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT5810HomilySatan" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Homily by Ephrem on Satan and on the death</title>
```
thea-m: en :)
"and on death" without the?

eu-genia: corrected (copypasted from E-S but you are right)

thea-m: sorry, my comment was maybe too elliptic. The title given here is English but marked up as geez.

eu-genia: ah:)

PR #13, opened 2019-11-19
Minor edits + new record

eu-genia:
Minor edits + new record

### DIFF
--- a/1001-2000/LIT1660Homily.xml
+++ b/1001-2000/LIT1660Homily.xml
@@ -11,7 +11,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
       <fileDesc>
          <titleStmt>
             <title xml:id="t1">Homily on the Passion of Our Lord</title>
-            <title xml:id="t2">Homily by Ephrem on Passion of Our Lord and on the thief on the right hand</title>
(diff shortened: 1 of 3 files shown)


1001-2000/LIT1660Homily.xml
```diff
@@ -11,6 +11,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
       <fileDesc>
          <titleStmt>
             <title xml:id="t1">Homily on the Passion of Our Lord</title>
+            <title xml:id="t2">Homily by Ephrem on Passion of Our Lord and on the thief on the right hand</title>
```
thea-m: "the" passion and "thief of", I would say

eu-genia: ok to the language correction

these are two different dionysius to my mind, this is why the change in PRS3587 to disambiguate

PRS3587: first century https://en.wikipedia.org/wiki/Dionysius_the_Areopagite
PRS7973: sixth century https://en.wikipedia.org/wiki/Pseudo-Dionysius_the_Areopagite

eu-genia: (but yes i should have inserted the other ID, corrected)

PR #27, opened 2020-01-22
Update LIT4275ChronAmdS.xml

sologebre:


PietroLiuzzo:
that was a small change :)

sologebre:
Ok, I will request at least one reviewer after I have made the correction to the text encoding. Then this will not happen again. Thank you very much!

PietroLiuzzo:
no worries!

PR #29, opened 2020-01-25
Update LIT2388Taamme.xml

PietroLiuzzo:
https://github.com/BetaMasaheft/Documentation/issues/1307

Here a SAWS class has been used. I cannot figure out what has happened so I just commented it out. the file history does not help me much to quickly figure out who to assign, I would invite some one to propose a solution, so that I can update this PR and merge the appropriate fix. 
@eu-genia @thea-m 

Thanks!


thea-m:
I think it can be deleted, since the relation to this general record is given in all the three records mentioned here.

PR #57, opened 2020-03-09
corrections

thea-m:


PietroLiuzzo:
sorry @thea-m I cannot reconnect these to any conversation (too many I am afraid... sigh!) However, I have the feeling that this lawd:hasAttestation, linking a work and a manuscript I understand that there is an unspecified passage of the manuscript which mentions this work (i.e. indirectly or by title). is this what this means? if not I think we need to specify it with a desc for the future. what do you think?

thea-m:
I'm sorry, I need to do a better job in general linking pull requests and issues...
Here, I looked at the last list of seven records mentioned in https://github.com/BetaMasaheft/Documentation/issues/1344 that were not reated elsewhere. 
As I understand the work-MS lawd:hasAttestation relations here, they are simply used to state that the work is contained in the manuscript. It is not something I would usually specify in a relation (because the work ID is used in the manuscript records), I kept it here not to lose the relation. Would it be better to delete it?

PietroLiuzzo:
I understand. is this what we did also for the miracles of mary?

thea-m:
Yes, that's it

PietroLiuzzo:
I will document this usage in the GL.

PR #63, opened 2020-03-10
split relations

thea-m:


thea-m:
merging to check whether this lead to an error on the app

PR #71, opened 2020-03-19
new record, corrected transliteration

thea-m:


thea-m:
added the reference, thank you, wonderful!

PR #73, opened 2020-03-23
Mazmura dengel

DariaElagina:
Added textparts for Mazmura Degel and the Song of Songs


thea-m:
thank you, very helpful! I just have one small question regarding the use of "Psalm" as value for @subtype for the Canticles and Song of Songs. Do you know if it used as terminus technicus for these elsewhere in the literature? It makes of course sense to have a common subtype for the entire work.

DariaElagina:
not really, that was my problem as well. Schould I rather you Canticle?

thea-m:
I'm not really sure, apparently because, as I notice now, I do not really understand the consequences of the use of @subtype. 
On the other hand, there are already subtypes as "PsalmoftheVirgin" in the record which are also restricted to it. So maybe "canticle" would be more precise. If you think psalm is better, just let me know and I will merge.

PR #87, opened 2020-04-02
added bibliography

thea-m:


2001-3000/LIT2701Dawit.xml
```diff
@@ -74,6 +75,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             
             <listBibl type="secondary">
                 <bibl><ptr target="bm:Heldman2010Psalter"/></bibl>
+                <bibl>
+                    <ptr target="bm:DegeMueller2015Psalter"/>
```
MassimoVilla, line 79: Add further bibliography

thea-m: Yes, there should absolutely be more references here. For training purposes, let us limit this PR to this one change. What do you think?

PR #94, opened 2020-04-09
Additions

thea-m:
Edits to titles, added bibliography
deleted LIT5087Sunday and LIT4918MalkeaMaryam (identical to LIT2889RepCh170), or did you think LIT5087Sunday might consitute an identifiable different recension? @DariaElagina 
created record for LIT5884IntroHymnPs - according to EMIP00040 it is the first psalm of Mazmura krəstos (which one?), but I don't have any edition or digitised manuscript available where I could check - does anyone have one? @DenisNosnitsin1970 @DariaElagina @eu-genia 

(not sure if the collecting edits in one PR approach makes things easier ;) )


thea-m:
@DenisNosnitsin1970

PR #97, opened 2020-04-12
Lit5843

DenisNosnitsin1970:
Full text of Lit 4859

PR #98, opened 2020-04-13
Create LIT5886MonthsLength.xml

DenisNosnitsin1970:
Calendaric note in DSEthiop2, to be upgraded

### DIFF
--- a/new/LIT5886MonthsLength.xml
+++ b/new/LIT5886MonthsLength.xml
@@ -1,0 +1,62 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT5886MonthsLength" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">MonthsLength</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DN"/>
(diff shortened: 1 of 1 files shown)


new/LIT5886MonthsLength.xml
```diff
@@ -0,0 +1,62 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT5886MonthsLength" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">MonthsLength</title>
```
thea-m, line 8: This is the title which will be displayed, could you give it the one you would like?

new/LIT5886MonthsLength.xml
```diff
@@ -0,0 +1,62 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="DN" when="2020-04-13">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT5886MonthsLength" passive=""/>
```
thea-m, line 57: Could you add the manuscript ID to @passive? But actually, the relation is not really necessary, the ID will be used in the manuscript record

new/LIT5886MonthsLength.xml
DenisNosnitsin1970: Done

new/LIT5886MonthsLength.xml
DenisNosnitsin1970: Done

thea-m:
I think you have carried out these changes in https://github.com/BetaMasaheft/Works/pull/97/files, so I will close this PR and we will only keep https://github.com/BetaMasaheft/Works/pull/97/files.

PR #103, opened 2020-04-20
updated work records

thea-m:


1001-2000/LIT1961Mashaf.xml
```diff
@@ -5,7 +5,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <fileDesc>
          <titleStmt>
             <title xml:lang="gez" xml:id="t1">መጽሐፈ፡ ሰዓታት፡</title>
-            <title xml:lang="gez" type="normalized" corresp="#t1">Maṣḥafa saʿātāt (general record)</title>
+            <title xml:lang="gez" type="normalized" corresp="#t1">Maṣḥafa saʿātāt</title>
+            <title xml:lang="en" type="normalized" corresp="#t1">Book of the hours</title>
```
DenisNosnitsin1970: I think we write the titles of the works with capital letters, don't we? Maṣḥafa saʿātāt "Book of the Hours", Sǝbḥata fǝqur "Praise of the Beloved". We do not do, if the title is descriptive and do not have an equivalent in Ethiopic, like: Lectionary for the days of the week.

PR #104, opened 2020-04-20
Lit1761 etc

DenisNosnitsin1970:
New record, additions


DenisNosnitsin1970:
If I record everything, in the records, it will cost additional time. If I am not akcnowledged in the app, it is ok, at least for now; i think the systems tracks all changes, doesn't is?

PR #117, opened 2020-04-27
Update LIT4960Taabbeyo.xml

DenisNosnitsin1970:
Text, titles, identification.

### DIFF
--- a/4001-5000/LIT4960Taabbeyo.xml
+++ b/4001-5000/LIT4960Taabbeyo.xml
@@ -10,6 +10,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                 <title xml:lang="gez" corresp="#t1" type="normalized">Saʾali lana Māryām saʾali lana Māryām ḫaba waldǝki ḫer madhane ʿālam</title>
                 <title xml:lang="en" corresp="#t1">Pray for us, Mary, Pray for us, Mary, to your Good Son Saviour of the World...</title>
 
+                <title xml:lang="gez" xml:id="t2">ተዐብዮ፡ ነፍስየ፡ ለእግዚአብሔር፡</title>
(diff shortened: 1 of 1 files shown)


4001-5000/LIT4960Taabbeyo.xml
```diff
@@ -1,14 +1,19 @@
...
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">ተዐብዮ፡ ነፍስየ፡ ለእግዚአብሔር፡</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Taʿabbǝyo nafsǝya la-ʾƎgziʾabǝḥer</title>
-                <title xml:lang="en" corresp="#t1">My soul magnifies God...</title>
+
+                <title xml:lang="gez" xml:id="t1">ሰአሊ፡ ለነ፡ ማርያም፡ ሰአሊ፡ ለነ፡ ማርያም፡ ኀበ፡ ወልድኪ፡ ኄር፡ መድኀኔ፡ ዓለም፡</title>
+
+                <title xml:lang="gez" corresp="#t1" type="normalized">Saʾali lana Māryām saʾali lana Māryām ḫaba waldǝki ḫer madhane ʿālam</title>
+                <title xml:lang="en" corresp="#t1">Pray for us, Mary, Pray for us, Mary, to your Good Son Saviour of the World...</title>
+
```
thea-m, line 12: I think it could be useful to keep ታዐብዮ፡ ነፍስየ፡ etc. as t2. Then it would not be printed in the article, but maybe people searching only for this expression would find this record easier.

DenisNosnitsin1970: OK, done. Actually in the HMML catalogue it seems to have been registered as "Sa'ali lana...". "Ta'abeyyo" is the initial words of the psalm required bebore the prayer.

PR #125, opened 2020-05-05
Create LIT5905FidalaHawarya.xml

DariaElagina:


new/LIT5905FidalaHawarya.xml
```diff
@@ -0,0 +1,72 @@
...
+        <body>
+            <div type="bibliography">
+              <listBibl type="secondary">
+                <bibl>
+                  <ptr target="bm:Platonov1996Katalog"/>
+                  <citedRange>26-29</citedRange>
+                </bibl>
+              </listBibl>
+                <listRelation>
+                    <relation name="saws:formsPartOf" active="LIT5905FidalaHawarya" passive="LIT3509Epistle"/>
```
eu-genia, line 67: saws:formsPartOf is used to declare that a defined Textual Unit is always part of another one.
I am not sure this is applicable here.
I would better put saws:isRelatedTo and add a desc element to explain the relation

PR #126, opened 2020-05-07
Edits DS

thea-m:


4001-5000/LIT4001Syllab.xml
```diff
@@ -4,7 +4,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title xml:id="t1">Syllabary of the Ethiopic alphabet</title>
```
DenisNosnitsin1970, line 7: I think we may need to diversify as there are two types of the syllabary (we dont do for the moment, do we?), one is Abugida-type and another is Hahu-type (EAE I, 55, anexcellent piece). In DS Ethiop 23, it is Hahu. I would suggest Ethiopic Syllabarium as the first title (I think "Ethiopic Abugida-syllabary" and "Ethiopic Hahu-syllabary" may work).

thea-m: You are right, I created it now in https://github.com/BetaMasaheft/Works/pull/128, so we can ignore it in this PR, it will be overwritten

thea-m:
Can this be merged? The syllabary issue is now treated separately

PR #137, opened 2020-05-14
Created salam hymn

MassimoVilla:


new/LIT5921SalamLaki.xml
```diff
@@ -0,0 +1,67 @@
...
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129i_is_subject_of" active="LOC2297DabraL" passive="LIT5921SalamLaki"/>
+                </listRelation>
+                <div type="edition">
+                    <div type="textpart" subtype="incipit" xml:lang="gez">
+                        <note>The incipit is taken from <ref type="mss" corresp="BDLaethe38" target="ms_a4"/>.</note>
+                        <ab>
+                            ሰላም፡ ለኪ፡ ደብረ፡ ሊባኖስ፡ ገነት፡ ፄና፡ መዓዛኪ፡ ዕፍረት፡ ዘአዕረፈ፡ ብኪ፡ ተክለ፡ ሃይማኖተ፡
```
SusanneHummel, line 60: everytig looks very nice!
Maybe ሃይማኖ!ት! ?

MassimoVilla: Thanks Sanne. That's how it looks like in the ms. BTW I noticed that the ms has ሃማኖተ፡, so I marked up the word as a scribal error `<choice><sic>ሃማኖተ፡</sic><corr>ሃይማኖት፡</corr></choice>`

new/LIT5921SalamLaki.xml
```diff
@@ -0,0 +1,67 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129i_is_subject_of" active="LOC2297DabraL" passive="LIT5921SalamLaki"/>
+                </listRelation>
+                <div type="edition">
+                    <div type="textpart" subtype="incipit" xml:lang="gez">
+                        <note>The incipit is taken from <ref type="mss" corresp="BDLaethe38" target="ms_a4"/>.</note>
```
thea-m, line 58: If you would like to refer to the precise xml:id of the item in the manuscript in ref, then you can use corresp="BDLaethe38#a4" (If this text is a4? otherwise the xml:id of whatever msItem or addition it is in the manuscript record). But referring only to the manuscript is also fine.

PR #140, opened 2020-05-15
Lit4959

DenisNosnitsin1970:
Minor change


DenisNosnitsin1970:
DS Ethiop 11, X), the title should be <title xml:lang="en" xml:id="t2" type="main"> "Homily on the Massacre of the Innocent and on the Magi for 3 Ṭərr, ascribed to John Chrysostom", but it is still "t1".

PR #150, opened 2020-05-25
Created lit5992 miracle

MassimoVilla:


new/LIT5992Miracle.xml
```diff
@@ -0,0 +1,68 @@
...
+        <revisionDesc>
+            <change who="MV" when="2020-05-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129i_is_subject_of" active="PRS5684JesusCh" passive="LIT5992Miracle"/>
+                    <relation name="ecrm:P129i_is_subject_of" active="PRS12983GabraMa" passive="LIT5992Miracle"/>
```
thea-m, line 54: is PRS12983GabraMa a protagonist of the miracle or is he just mentioned in the supplication?

new/LIT5992Miracle.xml
```diff
@@ -0,0 +1,68 @@
...
+                <keywords>
+                    <term key="Apocrypha"/>
+                    <term key="ChristianLiterature"/>
+                    <term key="Miracle"/>
+                </keywords>
+            </textClass>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="gez">Gǝʿǝz</language>
+            </langUsage>
```
thea-m, line 43: I'm sorry, I know it is annoying and takes time, but it would be helpful to have some information on the content to allow to disambiguate this miracle from others. Either in a very summary form in the title or in `<abstract>`. I think it would also improve the catalogue to have this kind of information

new/LIT5992Miracle.xml
MassimoVilla: Actually he's only mentioned, although it is not clear what is his role in the writing of the text. The text seems a quite recent creation rather than an excerpt from the well-known collection of "Taʾammǝra ʾIyasus".

PR #152, opened 2020-05-25
new record

thea-m:


new/LIT5993Hymn.xml
```diff
@@ -0,0 +1,67 @@
...
+            <change who="DR" when="2020-05-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="edition" xml:lang="gez">
+                <div type="textpart" subtype="incipit">
+                    <note>The following incipit is taken from <ref type="mss" corresp="BDLaethd20"/>.</note>
+                    <ab>
+                       ኦእግዝእትነ፡ ቅድስት፡ ድንግል፡ በ፪ማርያም፡ ማ<sic resp="DR">ረ</sic>ሃም፡ እሙ፡ ለኢየሱስ፡ ክርስቶስ፡ ከሀሉ፨
```
MassimoVilla, line 52: 1. Couldn't ከሀሉ፡ be ከሀሊ፡ (i.e., properly, ከሃሊ፡ "almighty") ?
2. In that case, the incipit looks like that of LIT3093RepCh373. The latter is found, among other mss, in ms Ethio-SPaRE DMB-003 (ESdmb003): I can see there on f. 9rb-va that both the incipit and the explicit are the same that the those here encoded (with some minor variants, such as ሃ in the name of Mary). I would be in favour of not creating this record and using LIT3093RepCh373 for the hymn in BDLaethd20. What do you think?

thea-m: You're right, it's the same. Wonderful, thank you!

PR #156, opened 2020-05-26
Update LIT3309Qerellos.xml

DariaElagina:
I have a question: I could add CPG IDs to the single textparts, do I do it inside listbibl with a corresp. Like that:`<listBibl type="clavis">
   <bibl corresp="#Dialogue" type="CPG"><ptr target="bm:CPG1"></ptr><citedRange unit="item">5228</citedRange></bibl>`
?

### DIFF
--- a/3001-4000/LIT3309Qerellos.xml
+++ b/3001-4000/LIT3309Qerellos.xml
@@ -11,10 +11,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <funder>Akademie der Wissenschaften in Hamburg</funder>
          </titleStmt>
          <editionStmt>
-          <p>Text structure id based on
-            <bibl>
-              <ptr target="bm:Bausi2010EAEQerellos"/>
-           </bibl>.</p>
+          <p></p>
          </editionStmt>
          <publicationStmt>
             <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
@@ -114,6 +111,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
               <relation name="saws:hasUsed" active="LIT3309Qerellos#VitaOfCyril" passive="LIT4723TarikaWaldaAmid"></relation>
             </listRelation>
             <div type="edition">
+              <note>Following text structure is based on
+                <bibl>
+                  <ptr target="bm:Bausi2010EAEQerellos"/>
+               </bibl>.</note>
 
               <div type="textpart" subtype="section" n="1" xml:id="Section1">
               <div type="textpart" subtype="treatise" xml:id="Prosphoneticus1">
--- a/3001-4000/LIT3309Qerellos.xml
+++ b/3001-4000/LIT3309Qerellos.xml
@@ -62,6 +62,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change who="DR" when="2017-02-14">Added keywords</change>
          <change who="MV" when="2017-12-27">Added witnesses, keywords, bibl</change>
          <change when="2020-05-26" who="DE">Added textparts and relations.</change>
+         <change when="2020-05-27" who="DE">Added corresps to textparts.</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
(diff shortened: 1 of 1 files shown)


eu-genia:
Well I wonder if a CPG ID is an indication that it should be a work record with its own ID

PietroLiuzzo:
> Well I wonder if a CPG ID is an indication that it should be a work record with its own ID
this would have to be decided on the specific case in question, isn't it? and following the GL on how to create . there is no support for this use of `@corresp` for the clavis bibliography. if it is needed, please open an issue.

DariaElagina:
Ok, and is there any, already existing way, to assign different CPG IDs to single textparts? I could not find any solution in the guidlines, unless I provide a bibliographical refernce inside each div, if it is allowed...

PietroLiuzzo:
the issue of the Textual Units needs to be sorted first, IMHO, thus identifying if there is such a need, that is cases of a Textual Unit which is never an independent TU but is so in e.g. the greek patristic. If this need is there you can already add the elements but there is no support for that, i.e. those associations will not surface anywhere in data or website, and I will need to build that feature in the application. If it is the case, as @eu-genia points out that these do need a TU, there is no need to develop this bit of software.

eu-genia:
Possibly @abausi will be able to tell whether (any of) the Qerellos homilies need independent IDs
(Also whether other Books of Qalementos, besides probably LIT1957Mashaf#Book5=LIT2680ClemPeter (?), need entries of their own?)

abausi:
I would be in favour of assigning IDs to parts corresponding to CPG; for this one can refer to my article on Qerǝllos in EAe, where precise correspondences are given

eu-genia:
Thank you, so as of now we do not need to assign CPG IDs to textparts, but we do need new work records

DariaElagina:
Ok, I can do it like that, but does all these parts circulate independently? Do we have evidence for that? Otherwise creation of separate first-level IDs would contradict our GL.

abausi:
I see the point; I do not remember exactly which is the situation for the single manuscripts of the "Qerǝllos" as described by Weischer, but this definitely happens for some of the sections. It is a situation that is parallel, for example, to what we have in the "Gadla samāʿtāt": if we have separate IDs for each piece within GS, even though definitely not each of them is attested separately, than we should have also IDs for each of the "Qerǝllos" section. I understand that this is an extremely delicate point, but I see all the disadvantages of not being able to highlight that single pieces are identified. I admit, I have not freshly read the GL on this point and there moght remain something important to rediscuss

DariaElagina:
Well, all sections and single identifiable units have their own second-level IDs, so that one can precisely refere to a specific part of that work in a manuscript description. This is not a problem, they are all identified. There is just no proper way to assign CPG IDs to the second Level IDs. But in any case I think that a proper documentation of the manuscript culture we are working should have priority over a Clavis designed for another written tradition. IMHO.

abausi:
Then this is a major point to discuss: to give up identification with CPG limits extremely the interoperability of our work; I remind of what emerged during Getatchew Haile's talk and exchange with Peter Brown: intercorrespondence between Greek tradition identified by CPG and Ethiopian and Eritrean manuscript culture is not at all a secondary point, quite the contrary, this is exactly why our work is well known also outside of the narrow limits of Ethiopian studies; this means that we have to find a good solution, because we cannot give up this essential point

DariaElagina:
I will create an issue then, of course, it is crucial to provide this information. @eu-genia , if everything else with the recod is ok, could it be merged like that, so that I can continue my work. Once we have a decision, I will come back to this record and edit it according to the decision we will agree upon.

eu-genia:
On rechecking: at least some, if not all, textparts already have independent BM IDs, so we need corresp elements for them

image

eu-genia:
Cp.
 https://betamasaheft.eu/manuscripts/BLadd16200/main
 https://betamasaheft.eu/manuscripts/IVkoriander3/main
 https://betamasaheft.eu/manuscripts/IVorlov4/main

abausi:
This is very good; actually, I remember I forwarded all CPG entries from EAe knwon to me and these were entered

DariaElagina:
But these are then not independent textual units, all these manuscripts have Qerellos. I mean of course, I just can make a corresp relation, but still this would be an encoding which contradicts the GL. These mss do not attest there independent circulation, rather circulation within the collection.

abausi:
I recall the parallel with the single GS texts: how were they noted? it is not exactly the same here, but it is very similar; not all GS units are positively attested outside the collection arrangement

DariaElagina:
Well, it might be that this principle is applied for other collections, the question is: On which principle the assigning of the first-level ID of Clavis should be built? On the evidence of independent circulation or on any other (not yet documented) principle. By now it might not be big issue, but once one would like to conduct a quantative or qualitive analysis on data, the difference between a textual unit with independet circulation and a textual unit attested only within a collection might be crucial.

abausi:
We must combine two clearly distinct facts: (1) to what extent a unit circulates independently; (2) how is this unit related with other traditions: there must be a way to consider both needs; moreover, we already have IDs for collections (e.g. the Miracles of Mary) and for the single units of the collections. I should re-read the GL on this and related points

eu-genia:
In order to not create too much confusion in the data as it is now, I would suggest to

(1) since the IDs already exist, with authors, bibliography, additional information provided - for now go on and add corresp elements to the div edition here (or remove back the edition and change the ref elements in the BLorient740 description, but at this point adding corresp is a more economic and logical solution) so that no information is lost

(2) should it be then decided that the homilies should be downgraded from text units to textparts, then at that point the record has to be thoroughly reedited again, incorporating all the data (CPG but also attributions and other bibliography) currently distributed among the 27 independent work records; in that workflow the records should be deleted (following the new procedure as here https://github.com/BetaMasaheft/Documentation/issues/1401) and all the MS records using them (the ones mentioned above + several EMIP records) have to be updated.

DariaElagina:
Ok, I will add the corresps then, so that at least no information is lost.

PietroLiuzzo:
As in the issue linked above I share my 2 cents: until 150% sure, deleting and "downgrading", both extremely meaningful actions, should be postponed with preference to an ID. documenting and enriching the guidelines with the existing examples would be great...  the new procedure will work from the release onwards, not right now, and only affects the consequences of deletion and the way in which to preserve the information.

eu-genia:
Sorry just noticed that one corresp is missing: TreatiseOnMelchizedek = LIT3332Melchiz

PR #161, opened 2020-05-28
Update LIT2285Saqoqa.xml

DenisNosnitsin1970:
Change in the title. Otherwise the pdf has two translations.


2001-3000/LIT2285Saqoqa.xml
```diff
@@ -7,7 +7,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <title xml:id="t1" xml:lang="gez">ሰቆቃወ፡ ድንግል፡</title>
             <title corresp="#t1" xml:lang="gez" type="normalized">Saqoqāwa dǝngǝl</title>
             <title corresp="#t1" xml:lang="en">Lament on the Virgin</title>
-            <title corresp="#t1" xml:lang="en">Elegy on the Virgin</title>
+            <title corresp="t2" xml:lang="en">Elegy on the Virgin</title>
```
eu-genia, line 10: I think it should have been
`xml:id="t2"`
(or it should have stayed `corresp="#t1"`)

DenisNosnitsin1970: Yes, sure.

PR #174, opened 2020-06-06
Update LIT3058RepCh338.xml

DenisNosnitsin1970:
Minor title change after EMML


3001-4000/LIT3058RepCh338.xml
```diff
@@ -8,7 +8,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <title xml:lang="gez" xml:id="t1">እሰግድ፡ ለኪ፡ እሰግድ፡ ለኪ፡ እሰግድ፡ ለኪ፡ ወእዌድሰኪ፡ </title>
             <title  xml:lang="gez" corresp="#t1" type="normalized">ʾƎsaggǝd laki ʾǝsaggǝd laki ʾǝsaggǝd laki wa-ʾǝweddǝsaki</title>
             <title  xml:lang="gez" corresp="#t1">ʾI worship you, I worship you, I worship you and I praise you...</title>
-            <title xml:lang="en" xml:id="t2" type="main">I worship Thee...</title>
+            <title xml:lang="en" xml:id="t2" type="main">I worship thee...</title>
```
thea-m, line 11: Here also, I would find it good if we had you/thee consistency over the catalogue. I would be for you, but if Aethiopica wants thee, we can take that...

PR #181, opened 2020-06-11
Litanies

thea-m:


thea-m:
The ሊጦን፡ ዘነግህ reference is kept in LIT4860Litanies, so I was hoping that people searching for this title would find the indidual records through this. It would also be correct in the individual litanies, which are part of this. I decided to delete it because it led to some wrong confusion when works are only designated as liṭon/kidān za-nagh in catalogues (without any text available) and random litanies could be by accident assigned. The references in the manuscripts to this title should be correct, because it might occur in manuscripts.

PR #212, opened 2020-07-01
Create LIT6047On12ApostlesRetuaH.xml

MarcinKrawczuk:


DariaElagina:
Dear @PietroLiuzzo , would such an ID be ok?

PietroLiuzzo:
should be, yes, it starts, after the numeri part with a letter. however, for future record creation, please, do have a look at the id assignment guidelines https://betamasaheft.eu/Guidelines/?id=entities-id-structure one word should suffice.

PR #215, opened 2020-07-02
new record

thea-m:


DariaElagina:
Does it have an independent circulation? Otherwise it can be just added as div

thea-m:
Since it was specifically mentioned only in one manuscript, it does not seem to part of the standard of this work, and then fullfils the criterion of possible absence for an independent ID.

PR #223, opened 2020-07-07
New entries for b ladd19341

DariaElagina:


thea-m:
It is also attested in EMIP00706 and EMIP02082 as part of "groups of (school) chants", maybe this term and the keyword Poetry (and maybe DidacticMaterial?) could also be used?

PR #225, opened 2020-07-07
IDcorrection

eu-genia:
renamed + deleted


eu-genia:
sorry @PietroLiuzzo  there is some conflict here I guess because of some overlapping that I cannot resolve 
can you help please...

PietroLiuzzo:
perhaps you forgot to update master so that it had your[ previously merged changes](https://github.com/BetaMasaheft/Works/pull/224/files), before doing this correction? I think this pr and branch can be deleted, then a new branch and pr starting from the latest master could be done.
next time, I think it would be wiser to simply do both file renamaing and updating of contents in one go, they really do not need two PRs

PR #234, opened 2020-07-08
Collectio monastica

DariaElagina:


new/LIT6074TwelveFathers.xml
```diff
@@ -0,0 +1,83 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+
+            <div type="bibliography">
+              <listRelation>
+                <relation name="saws:formsPartOf" active="LIT6074TwelveFathers" passive="LIT4838CollectioM"></relation>
+                <relation name="saws:isAttributedToAuthor" active="LIT6074TwelveFathers" passive="PRS5585Isaiaht">
+                  <desc>The Syriac version of this text is attributed to <persName ref="PRS5585Isaiaht"></persName></desc>
```
thea-m, line 62: I think you could give a secondary reference to wright and the EAE for this record (and the next)

4001-5000/LIT4838CollectioM.xml
```diff
@@ -85,8 +86,142 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <ab>
                  ቀዳሜ፡ ትእዛዙ፡ ለአብ፡ ዘይቤ፡ ዘቦ፡ እዝነ፡ ለይስማዕ። ይቤ፡ እግዚአብሔር፡ ሰማይ፡ መንበርየ፡
                </ab>
+               <div type="textpart" subtype="chapter" n="1" xml:id="Chapter1" xml:lang="gez">
```
thea-m, line 89: I assume that the textparts follow the edition, but it would maybe be better to clarify this

PR #241, opened 2020-07-17
Emip new works 17 july 2020

Ralph-Lee-UK:
Final set of new work files for the EMIP project


PietroLiuzzo:
I will merge this, but I think Works are a rather sensitive issue, and it may be better to assign one of my colleagues as a reviewer in the future.

Ralph-Lee-UK:
Thanks for this, Pietro. I would agree that this is difficult, and indeed some of the 'works' in the list that I had are clearly not. I will be sure to run any that I have doubts over past Dorothea and others - the community is working very well for me these days, with good discussions, clear decisions, great feedback!

PR #243, opened 2020-07-19
Update LIT1970Mashaf.xml

DenisNosnitsin1970:
Some updates


thea-m:
The list of witnesses is not necessary, unless you are planning an edition using these manuscripts? 
If you want to list all witnesses, because the ID is not mentioned in the manuscript records, it would be better to work with the relation lawd:hasAttestation.

PR #244, opened 2020-07-19
Update LIT1970Mashaf.xml

DenisNosnitsin1970:
New witnesses


DariaElagina:
You mean that they are going to be used in the digital edition?

DariaElagina:
Are those changes equal to those in the previous PR?

DenisNosnitsin1970:
Maybe. What previous PR do you mean?

DariaElagina:
According to the guideleines: We use <witness>↗ for those manuscripts which are used for the edition we are going to provide in the <div type='edition'>↗. 
I mean, #243

PR #245, opened 2020-07-20
Create LIT6156OriginProphets.xml

DenisNosnitsin1970:
New record

### DIFF
--- a/new/LIT6156OriginProphets.xml
+++ b/new/LIT6156OriginProphets.xml
@@ -61,9 +61,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </listRelation>
                    </div>
                 <div type="edition">
-                  <note>The division into text parts follow <ref type="mss" corresp="BNFabb195"/>.</note>
+                  <note>The division into text parts follows <ref type="mss" corresp="BNFabb195"/>.</note>
                    <div type="textpart" subtype="chapter" xml:id="Chronology">
-                      <label>On the Chronology of the Prophets</label>
+                      <label>On the chronology of the Prophets</label>
                    </div>
                    <div type="textpart" subtype="chapter" xml:id="Isaiah">
                       <label>On Prophet Isaiah</label>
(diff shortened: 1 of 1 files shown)


new/LIT6156OriginProphets.xml
```diff
@@ -0,0 +1,100 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="DN" when="2020-07-20">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT6156OriginProphets" passive="BNFabb195"/>
```
DariaElagina, line 60: If you will provide ID of the work in the manuscript description, this information is actually superfluous

new/LIT6156OriginProphets.xml
```diff
@@ -0,0 +1,100 @@
...
+                </listRelation>
+                   </div>
+                <div type="edition">
+                  <note>The division into text parts follow <ref type="mss" corresp="BNFabb195"/>.</note>
```
DariaElagina, line 64: follows

DariaElagina:
Please, note that we have agreed upon creating an issue first before creating a new work record to clear if a new record is really needed

new/LIT6156OriginProphets.xml
DenisNosnitsin1970: I alwazs provide at least one, otherwise the system reports a mistake?

new/LIT6156OriginProphets.xml
DenisNosnitsin1970: ok

DenisNosnitsin1970:
Ok, here the result will be 99% the same.

new/LIT6156OriginProphets.xml
DariaElagina: no relation is mandatory, there should be no mistake

PR #265, opened 2020-08-07
Create LIT6166MiraclesJohnBaptist.xml

DenisNosnitsin1970:
New entry, attested in a small number of mss.


PietroLiuzzo:
does this follow on a discussion on creating the record? perhaps it should be linked. The placeholder file created on the app has been publicly available without sync to this record for some time. Please review and merge asap, so that eventually if it was not really needed this record can also be removed from the db. thanks!

PR #281, opened 2020-09-07
Update title(s), witnesses, and bibliography

smaugustine:
A critical edition and translation of LS is an ongoing project of mine (and NB that it is the main content of BAVet 208, which I shall add shortly).

The relevant bibliography (missing from Zotero) would be the entry in the EAe (553b-554b) and also:
- Grébaut, Sylvain. “L’Hymne-Invocation Lesâna Sab’e.” _Aethiopica : revue philologique_ 3, no. 1 (1935): 6–12.


PietroLiuzzo:
please for request about the bibliography, use this issue https://github.com/BetaMasaheft/Documentation/issues/1445

eu-genia:
For Grébaut you can use bm:Grebaut1935LesanaSab
For the EAe entry (which was in Zotero, just without bm tag) you can use bm:HabtemichaelNosnitsin2007Lessanasab

For the future if you need a new record you should use this issue
https://github.com/BetaMasaheft/Documentation/issues/1445

Thank you

PR #289, opened 2020-09-15
Gadla Sadqan zaBaraknaha (and a small update to Sayfa Sellase)

eu-genia:


new/LIT6176MalkaSadeqanB.xml
```diff
@@ -0,0 +1,65 @@
...
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="ES" when="2020-09-14">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listBibl type="text">
```
thea-m, line 55: I still do not understand the use of listBibl type="text", I will be happy to enlightened on this.

1001-2000/LIT1489Gadlas.xml
```diff
@@ -26,6 +26,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <p>Work of the literatures of Ethiopia and Eritrea</p>
          </sourceDesc>
          <sourceDesc>
+            <listWit>
+               
+               <witness corresp="BaraknahaSadqan">Ms. from the church of Bāraknāhā (Eritrea) (publ. <bibl>
```
thea-m, line 31: According to the guidelines, `witness `is only used for manuscripts used for the edition provided by the record. I understand the need to save all the witness information given in the Ethio-SPaRe records, of which many manuscripts will not be catalogued in BM in a long time, which would be a pity to lose in the meanwhile.
But I don't think that giving non-existent IDs in `@corresp` is the best way to go. I wonder whether this information might be better recorded in `listBibl` (secondary or a new attribute for this) , if we stay with the current guidelines on witnesses.
This is a common issue with the updating of Ethio-SPaRe records, if the encoding as in this example has already been discussed and I'm not aware of it, the guidelines could be updated. @eu-genia @PietroLiuzzo

thea-m (changes requested):
thank you, just two questions

1001-2000/LIT1489Gadlas.xml
eu-genia: they are existent https://github.com/BetaMasaheft/Manuscripts/pull/767

eu-genia: I created stubs for all MSS

new/LIT6176MalkaSadeqanB.xml
eu-genia: the text is reproduced in the article (just checked) but it is surely not an edition

thea-m (approved):

PR #298, opened 2020-09-18
Correction to the colophon in the Miracles of George

eu-genia:


thea-m:
I think you might have taken the text from abbyy without correcting it? :)

thea-m (approved):

PR #302, opened 2020-09-21
Homily churches

MarcinKrawczuk:
Here is the record where the title has been taken from another ms but I didn't know how to link to it.

### DIFF
--- a/new/LIT6180HomilyChurches.xml
+++ b/new/LIT6180HomilyChurches.xml
@@ -1,0 +1,63 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6180HomilyChurches" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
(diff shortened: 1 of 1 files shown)


new/LIT6180HomilyChurches.xml
```diff
@@ -0,0 +1,64 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>The title of the work has been taken from EMML 8897, <locus target="#49r">f. 49r</locus>.</p>
```
DariaElagina, line 36: Since there is a record for the ms you can add a <ref> pointing to this particular ms: `<ref type="mss" corresp="..."></ref>`

new/LIT6180HomilyChurches.xml
```diff
@@ -0,0 +1,64 @@
...
+        <revisionDesc>
+            <change who="MKr" when="2020-09-15">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129_is_about" active="LIT6180HomilyChurches" passive="PRS6819Mary"/>
+                    <relation name="saws:isDifferentTo" active="LIT6180HomilyChurches" passive="LIT6178BasilPhilippi"/>
```
DariaElagina, line 59: The attribution to an author can also be encoded here (saws:isAttributedToAuthor), with a precise description and a ref to the Ms

MarcinKrawczuk: Thank you for that. In EMML 8897 it is attributed to John of Cesarea, my intuition is that they made a mistake and it should be Basil of C., but it would take some research to confirm. So for the time being I ignored the attribution.

DariaElagina (approved):

PR #321, opened 2020-10-10
Create LIT6193HomiliaryPassions.xml

DenisNosnitsin1970:
New record, to dissimilate from LIT1291Dersan

### DIFF
--- a/new/LIT6193HomiliaryPassions.xml
+++ b/new/LIT6193HomiliaryPassions.xml
@@ -1,0 +1,155 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6193HomiliaryPassions" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Homiliary on the Passions of Christ</title>
+                <title xml:lang="gez" xml:id="t2">ድርሳን፡ ማኅያዊ፡</title>
+
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DN"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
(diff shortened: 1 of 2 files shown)


new/LIT6193HomiliaryPassions.xml
```diff
@@ -0,0 +1,155 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>The work is different from <ref type="work" corresp="LIT1291Dersan"/>.</p>
```
eu-genia: relation can be created for this; the record LIT1291Dersan should also be corrected accordingly (it is written there 

```
It is not entirely clear from the catalogues whether the work contained in
<ref type="mss" corresp="BNFet116#ms_i1"/>
is the same as in
<ref type="mss" corresp="Bladd16254#ms_i1"/>
and
<ref type="mss" corresp="BNFabb29"/>
.
```

new/LIT6193HomiliaryPassions.xml
```diff
@@ -0,0 +1,155 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
```
eu-genia, line 1: was there an issue opened to discuss the new record? did I miss it?

new/LIT6193HomiliaryPassions.xml
```diff
@@ -0,0 +1,155 @@
...
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6193HomiliaryPassions" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Homiliary on the Passions of Christ</title>
+                <title xml:lang="gez" xml:id="t2">ድርሳን፡ ማኅያዊ፡</title>
+
```
eu-genia, line 12: transliterated title?

new/LIT6193HomiliaryPassions.xml
```diff
@@ -0,0 +1,155 @@
...
+          <ab>
+            <pb n="36r"></pb>
+            <cb n="a"></cb>ለዘጸሐፎ፡ ወለዘአጽሐፎ፡ ለዝንቱ፡ ድርሳን፡ እግዚአብሔር፡ ይጽሐፍ፡ ስሞ፡ በአምደ፡<cb n="a"></cb>
+             ብርሃን፡
+          </ab>
+          </div>
+          </div>
+
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT6193HomiliaryPassions" passive="BNFabb29"/>
```
eu-genia, line 144: this belongs to the listWit not relation

eu-genia (changes requested):

new/LIT6193HomiliaryPassions.xml
DenisNosnitsin1970: Yes, there was an issue, some time ago, initiated by Ralph.

new/LIT6193HomiliaryPassions.xml
DenisNosnitsin1970: Ok

new/LIT6193HomiliaryPassions.xml
DenisNosnitsin1970: They are different. Ok, I'll clarify.

new/LIT6193HomiliaryPassions.xml
DenisNosnitsin1970: Someone told me, in a very certain way, that listWit is used only for (digital) edition. I always included witness mss into this field, i think the others as well.

DenisNosnitsin1970:
I did some small changes in the records for work. The MS in Dillmann, BL, cannot be checked for the moment. I will ask Ralph to correct EMIP ms using the new ID.

eu-genia (approved):

new/LIT6193HomiliaryPassions.xml
thea-m: listWit should only be used for manuscripts used for an edition published in the same record. As there were different guidelines at the beginning of BM, there is no need to update the old records, but this should not be replicated. However, I would also not use the hasAttestation relation here (though not wrong, no need to go back and change the record now) - as long as the work ID is referenced in the manuscript, it will be picked up by the app.

PR #332, opened 2020-10-20
Update LIT5188MashafaMadbal.xml

DariaElagina:


5001-6000/LIT5188MashafaMadbal.xml
```diff
@@ -1,20 +1,20 @@
...
 type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT5188MashafaMadbal" xml:lang="en" type="work">
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">መጽሐፈ፡ መድበል፡</title>
                 <title xml:lang="gez" corresp="#t1" type="normalized">Maṣḥafa Madbal</title>
-                
+
                 <title xml:lang="gez" xml:id="t2">ጉባዔ፡ ምሥጢራተ፡ ዘመናት፡</title>
                 <title xml:lang="gez" corresp="#t2" type="normalized">Gubāʿe Mǝśṭirata Zamanāt</title>
```
thea-m: Our guidelines only state the capitalising of the first word of titles, is there a particular reason for the capitalising throughout?

5001-6000/LIT5188MashafaMadbal.xml
```diff
@@ -53,22 +53,67 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                    <label xml:lang="en">Introduction to the second part</label>
+                  </div>
+                  <div type="textpart" subtype="chapter" xml:lang="gez" xml:id="Verses">
+                    <label xml:lang="en">Verses addresses to the Holy Trinity</label>
+                  </div>
+                  <div type="textpart" subtype="chapter" xml:lang="gez" xml:id="FourGospels">
+                    <label xml:lang="en">The four Gospels in verses</label>
+                  </div>
+                  <div type="textpart" subtype="chapter" xml:lang="gez" xml:id="Testament">
+                    <label xml:lang="en"><foreign xml:lang="gez">መጽሐፈ፡ ኪዳን፡</foreign>in verses</label>
```
thea-m: I would use here a ref with the work IDs, or otherwise, transcribe the titles for better ease of reading

thea-m (changes requested):
Thank you!

5001-6000/LIT5188MashafaMadbal.xml
DariaElagina: no, I just have forgotten to check the titles

5001-6000/LIT5188MashafaMadbal.xml
DariaElagina: The refs are not appropriate here, as the texts appear to be rather retelling of other texts

thea-m: (It does not matter in this case, but it is my understanding that the refs are used to point to the works in cases like this, they do not state the present of the works)

thea-m (approved):

PR #338, opened 2020-10-21
Update LIT4668Prayer (Prayer of Peter)

smaugustine:
- Update the various titles, which were awkwardly literal translations of Grébaut's Latin
- Update the primary title, ጸሎቱ፡ ለጴጥሮስ፡, which is given within the incipit and used in publications related to the text (and in other manuscript catalogues)
- Add incipits taken from the published book/critical edition (which are essentially the same text)
- Add bibliography
- Change BAVet20 (which is a Psalter) to BAVet92

It's perhaps the case that `#Andrew` should be grouped with the prayers for Sunday and `#Enoch` be grouped with the daily prayers, since that's the division suggested by the printed text (as opposed to Grébaut's artificial sectioning and titles), but I wasn't sure if others had an opinion on this.

### DIFF
--- a/4001-5000/LIT4668Prayer.xml
+++ b/4001-5000/LIT4668Prayer.xml
@@ -164,7 +164,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
           </div>
 
           <div type="textpart" xml:id="Saturday">
-            <label>Prayer for Saturday</label>
+            <label>Prayers for Saturday</label>
 
             <div type="textpart" xml:id="Saturday1">
               <label>First Prayer for Saturday</label>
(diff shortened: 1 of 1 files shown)


eu-genia (approved):

eu-genia:
@DariaElagina followed the grouping in RNBspbdaBIII16 so she should have a say on this; in any case there is no need to suppress the labels, one could modify the nesting if necessary

DariaElagina:
Thank you! I think that if this record is meant to represent the existing critical edition, it should follow the edition. The description of the manuscripts might have another nesting. The manuscript RNBspbdaBIII16 seems to me to have other texts as well (I should however take a closer look).

PR #344, opened 2020-10-29
Lit6208 cyriacus pilate

MarcinKrawczuk:
The title is on p. 161 (no. 10) of Wright's catalogue, I have abbreviated it slightly, I'm not sure whether this is acceptable.


thea-m (approved):

thea-m:
Thank you! Just a quick comment for the future: It is always a good practice to add a link to the issue in which the creation of the record was discussed in the PR.

PR #354, opened 2020-11-09
Update LIT1239Chroni.xml

DariaElagina:
Dear Pietro, would that be a ligitimate way to link the text to the external parallel tradition?

### DIFF
--- a/1001-2000/LIT1239Chroni.xml
+++ b/1001-2000/LIT1239Chroni.xml
@@ -129,7 +129,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <listRelation>
                <relation name="skos:broadMatch"
                          active="https://malalas.hadw-bw.de/kommentar/1/5"
-                         passive="betmas:LIT1239Chroni#Chapter4">
+                         passive="betmas:LIT1239Chroni.4">
                          <desc>Parallel transmission in John Malalas 1.5.</desc>
                        </relation>
               </listRelation>
--- a/1001-2000/LIT1239Chroni.xml
+++ b/1001-2000/LIT1239Chroni.xml
@@ -4,12 +4,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title xml:lang="gez" xml:id="t1">ዮሐንስ፡ መደበር፡</title>
+            <title xml:lang="en" xml:id="t1" type="main"> Chronicle of John of Nikiu</title>
                <title xml:lang="gez" xml:id="t2">ዮሐንስ፡ መደብር፡</title>
-                 <title xml:lang="gez" corresp="#t2">Yoḥannǝs Madabbǝr</title>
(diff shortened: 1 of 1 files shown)


eu-genia:
I am not sure that this is the way we treat these things. Is it?
Especially having an external source as active?
Do we have guidelines for this?

What I see in the Guidelines for parallel versions (Ok same language but still)
If you have multiple recensions/version of a text, these can be linked with a `<relation>`↗ with `<name='saws:isVersionOf'>`↗ this will be used to pick up these in the text view. The same parallel content in the different versions should have the same `@corresp` to a NAR unit and there should always be a `@resp` in the `<div[@type='edition']>`↗ containing this text part with multiple versions with either the initials of the editor or the pointer to the @xml:id of the bibl element which has a pointer to the edition.

So to me it looks like the narrative, if it is shared by several texts, should get a NAR type ID, where this NAR can have then the description, that it appears in these two contexts, and here from Chapter 4 you add a corresp to the NAR ID.

But maybe I read it all wrong.

By the way, there are things to fix in the title area
```

            <title xml:lang="gez" xml:id="t1">ዮሐንስ፡ መደበር፡</title>
        REMOVE       <title xml:lang="gez" xml:id="t2">ዮሐንስ፡ መደብር፡</title>
       REMOVE          <title xml:lang="gez" corresp="#t2">Yoḥannǝs Madabbǝr</title>
            <title xml:lang="gez" corresp="#t1"  **type="normalized"**>Yoḥannǝs Madabbar</title>
            <title xml:lang="en" corresp="#t1" **type="main"**>Chronicle of John of Nikiu</title>
     REMOVE:       <author>John of Nikiu</author>

```

DariaElagina:
Dear Zhenia,

many thanks to your feedback. I was following this instruction in the GL:
External entries in databases or similar which concern a particular work can also be referenced in a <relation>↗. If, as in the case of the TUK IDs given by Corpus Coranicum, the external entry refers only to a part of a work, it is preferable to add the <relation>↗ in the exact <div>↗ of this text part in our edition, using <relation name="skos:exactMatch">↗. If this is not possible, because we do not yet have the text of the work or because we follow different divisions than the other reference, <relation name="skos:broadMatch">↗ can be used at the most pertinent place of the record with the most precise pointer to our record possible in @passive, as, for example, in


      <listRelation>
      <relation name="skos:broadMatch" active="https://corpuscoranicum.de/kontexte/index/sure/6/vers/74/intertext/673" passive="LIT1697Jubilees"></relation>
      </listRelation>
[https://betamasaheft.eu/Guidelines/?id=work-teiHeader](url)

I thought that it might be an economic and practical way to encode any relation to the external entity, as defined also in the GL:

skos:broadMatch | Fine for everything, especially to match with external entities.
-- | --

eu-genia:
You are right, this seems the way.
Should it be betmas:LIT1239Chroni.4 and not, as in the guidelines example, betmas:LIT1239Chroni:4 (colon?)

PietroLiuzzo:
@eu-genia is right, I am fixing that example straight away. 
here is the page of the guidelines with referencing instructions
https://betamasaheft.eu/Guidelines/?q=references&start=16&id=referencing-the-text

DariaElagina:
so, ist betmas:LIT1239Chroni.4  right? As far as I can see from the guidelines?

eu-genia (approved):

PR #355, opened 2020-11-10
Miracles mary

thea-m:


DariaElagina:
Thank you! An impressive work! Do I understand correctly that some collections of miracles have introdiuctory texts and others don't?

DenisNosnitsin1970 (approved):

DenisNosnitsin1970 (approved):

thea-m:
Thank you! Yes, LIT2384Taamme is for introductions and miracles, for what we would traditionally identify as Miracles of Mary. LIT3585Taamme is for more than one miracle occurring together, so it is also a part of LIT2384Taamme.

DariaElagina (approved):

PR #362, opened 2020-11-13
Added full text and relation in short hymns

SusanneHummel:
Here comes the PR as annoinced in the issue: https://github.com/BetaMasaheft/Documentation/issues/1557


2001-3000/LIT2988RepCh269.xml
```diff
@@ -10,63 +11,77 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+         <change who="PL" when="2016-07-26">Created XML record from Ethio authority google
+            spreadsheet</change>
          <change who="MV" when="2017-06-23">Added keywords and incipit</change>
+         <change who="SH" when="2020-11-13">Added full text and relation</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
+
+         <listRelation>
```
thea-m, line 58: These relations are not necessary (not wrong either, but redundant, because the ID is used in the manuscript and you give the source for your text).

3001-4000/LIT3002RepCh282.xml
```diff
@@ -10,63 +11,75 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
          <change who="MV" when="2017-02-16">Added keywords and incipit</change>
+         <change who="SH" when="2020-11-13">Added full text and relation</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
+
+         <listRelation>
+            <relation name="lawd:hasAttestation" active="LIT2993RepCh274" passive="MNC010"/>
+         </listRelation>
```
DariaElagina, line 60: it should be LIT3002RepCh282

thea-m (approved):

2001-3000/LIT2988RepCh269.xml
SusanneHummel: Okay, I can delete them. I added them in accordance to the Miracles of George. Howver, no problem. I can delete them

DariaElagina:
Thank you! I am not sure that an incipit as a text part is needed. At least I have never done like that by mysel, but maybe I was doing it wrong...

SusanneHummel:
I just added the full text for clarity since the hymns are so short. This might help to prevent confusion (issue: #1557)
The incipits were already given.

thea-m:
Actually, I think @DariaElagina has a point, the double incipit might be confusing and is unnecessary. I will delete them.

PR #384, opened 2020-11-24
Update LIT4957SebhataF.xml

DenisNosnitsin1970:
Incipit of Sebhate fequr zaGiyorgis


SusanneHummel (commented):
Sorry, I don't understand from where this incipit comes from. Was this text already in the work record und you just changed it? https://betamasaheft.eu/works/LIT4957SebhataF/main 
However, maybe one should also add the explicit?

DenisNosnitsin1970:
The text is taken from the edition mentioned in the record already (Mogas Equba Giyorgis); a couple of line from the explicit is quoted just below the incipit. When the PR is approved, the entire information will be visible under LIT4957SebhataF.

SusanneHummel:
Sorry, I most probably just checked your last commit...

DenisNosnitsin1970:
You can see the entire record with bibliography if you click on tiny signs in the upper and bottom left corner (arrows, option "expand") when you look and changes under "files changed". If you will create a record for the prayer of which you said it is dissimilar of "Sebhate fequr" but share only the couple of first lines, you can add "dissimilar to" in the relations.

SusanneHummel:
Do you have a pdf of this edition? I would like to check the explicit for comparison with the text attested in MNC-011. the beginning is exactly the same, the explicit is different. Maybe also a bit short for proper comparison?

Again, this is a complete different text than the text attested in MNC-008  and in the other edition of the Saʿātāt as disussed in the issue https://github.com/BetaMasaheft/Documentation/issues/1572

SusanneHummel (approved):

PR #394, opened 2020-11-30
Update LIT1191Baenta.xml

thea-m:
@DenisNosnitsin1970 , this is relevant for the DS catalogue


DenisNosnitsin1970 (approved):
This is the standard translation, i think.

PR #410, opened 2020-12-05
Lit1664 homily corr mk

MarcinKrawczuk:


eu-genia (approved):
Just for the future - make sure that you create new branches for new changes, from Master, to avoid confusions. Thank you

PR #416, opened 2020-12-09
Update LIT4015Mashafa.xml

thea-m:


DenisNosnitsin1970 (changes requested):
Just a comment: I find the title assigned to the work, Mashafa temqät, unlucky because it strongly interferes with the very common Mashafa temqat = Mashafa krestenna. Unfortunately, there is no way to consult Vat et 4; can we change the title of LIT4015 (Mashafa temqat may stand as the second title, if you like)? Lets say, it might be "Liturgy for he Feast of Epiphany" (it is not a homiliary, in the best case it is a kind of divine office).

DariaElagina (approved):

PR #432, opened 2020-12-17
Lit1922 mashaf corr mk

MarcinKrawczuk:
There might be some connection between መጽሐፈ ዶርሆ and LIT1274Dersan, but I'm not acquainted with this literature, just wanted to add bibliography.


eu-genia (approved):

eu-genia:
have you tried removing and recloning your works directory and restarting from master? (or ask pietro what the problem is, i remember augstine too had a similar case, don't remember the solution...)

PR #441, opened 2021-01-06
Lit2222 retua h corr mk

MarcinKrawczuk:
Hello in the New Year :) To get it started I have expanded the record of my favourite collection as much as I could. If you have any suggestions that may be useful for the users, they are most welcome.


2001-3000/LIT2222RetuaH.xml
```diff
@@ -42,8 +59,114 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+<div type="textpart" subtype="homily" xml:lang="gez" xml:id="Sunday2" corresp="LIT6050SundayRetuaH">
+<label>On fasting and peace, for the second Sunday of Lent</label>
+</div>
+
+<div type="textpart" subtype="homily" xml:lang="gez" xml:id="Sunday3" corresp="LIT6061SundayRetuaH">
+<label>On fasting and meekness, for the third Sunday of Lent</label>
+</div>
+
+<div type="textpart" subtype="homily" xml:lang="gez" xml:id="Sunday4">
+</div>
```
thea-m, line 114: I think that labels here (even minimal, "For the fourth Sunday" or similar, would be helpful for readers.

2001-3000/LIT2222RetuaH.xml
```diff
@@ -33,7 +34,23 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+            <keywords>
+               <term key="ChristianLiterature"/>
+               <term key="Homily"/>
+            </keywords>
+         </textClass>
+<abstract>
+<p>Rǝtuʿa Hāymānot is a collection of homilies focusing on the events of the Passion cycle. Most probably it is an indigenous
+  Ethiopian production. Its as author must have lived ca. 1375 and there is some ground to identify him
+  with <persName ref="PRS4709Giyorgis">Giyorgis of Saglā</persName>. It should not be confused with a collection of homilies
+preserved in <ref type="mss" corresp="EMML1763">EMML 1763</ref>, <ref type="mss" corresp="EMML8509">EMML 8509</ref>
```
thea-m, line 51: If there would be an ID for this other collection, the users would be happy to find it, but I assume it doesn't exist yet.

thea-m (changes requested):
Happy new year! It looks great, just a minor remark on labels.

2001-3000/LIT2222RetuaH.xml
MarcinKrawczuk: No it doesn't. I hope at one point to distinguish between various persons called RH and collections.

thea-m (approved):

PR #448, opened 2021-01-18
Lit5859 homily blind corr mk

MarcinKrawczuk:
I've added a bibliographical reference, it doesn't mention Ethiopic tradition but it surely be interesting to anyone who deals with this text.


eu-genia (approved):

eu-genia:
maybe @PietroLiuzzo can help you with cleaning the cash

PR #451, opened 2021-01-19
Update LIT3177Canon.xml

Ralph-Lee-UK:
Added further titles and witnesses. In discussion with Denis we identified three witnesses to this work, and I have also supplied an alternative English title, as well as a Ge'ez title

### DIFF
--- a/3001-4000/LIT3177Canon.xml
+++ b/3001-4000/LIT3177Canon.xml
@@ -4,9 +4,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title xml:id="t1">Canon of the Doctors of the Church on those who abjure</title>
-            <title xml:id="t2">Canon Established by of the Doctors of the Church on those who deny the faith</title>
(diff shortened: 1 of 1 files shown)


3001-4000/LIT3177Canon.xml
```diff
@@ -25,7 +28,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </sourceDesc>
          <sourceDesc>
             <listWit>
-               <witness/>
+              <witness corresp="EMIP02264#ms_i3"/>
```
thea-m, line 31: This encoding of witnesses does not correspond (anymore) to the guidelines: only in case of an edition present in this file should the witnesses used for this edition be encoded here (https://betamasaheft.eu/Guidelines/?id=work-teiHeader).
If these manuscripts will not be described anythime soon and the work ID therefore not be used in their description and you want to keep the information that they contain this work, you could add the relation lawd:hasAttestation.

thea-m (changes requested):
Thank you!

Ralph-Lee-UK (commented):
Thanks for spotting this

3001-4000/LIT3177Canon.xml
Ralph-Lee-UK: I think that I have done this correctly - please let me know if there are any further errors.

Ralph-Lee-UK:
@thea-m when I pull this latest edit and look at the file in Atom there is still an error - but I can't see how to correct it. Can you advise?

thea-m:
Good that you spotted it! It was just the deleted `<ab/>` missing, I added it back.

thea-m (approved):

PR #457, opened 2021-01-25
new ID, updates

thea-m:
@Ralph-Lee-UK 
https://github.com/BetaMasaheft/Documentation/issues/1672

### DIFF
--- a/4001-5000/LIT4551Salam.xml
+++ b/4001-5000/LIT4551Salam.xml
@@ -2,7 +2,7 @@
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title xml:lang="en" xml:id="t1">Salām-hymn to Lālibalā</title>
+                <title xml:lang="en" xml:id="t1">Hymn to Lālibalā</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="DR"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -19,11 +19,6 @@
             <sourceDesc>
             <p>Work of the literatures of Ethiopia and Eritrea</p>
          </sourceDesc>
-         <sourceDesc>
-                <listWit>
-                    <witness corresp="BDLaethe4#ms_i5"/>
-                </listWit>
-            </sourceDesc>
         </fileDesc>
         <encodingDesc>
             <p>A digital born TEI file</p>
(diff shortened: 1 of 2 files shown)


4001-5000/LIT4551Salam.xml
```diff
@@ -2,7 +2,7 @@
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title xml:lang="en" xml:id="t1">Salām-hymn to Lālibalā</title>
```
DenisNosnitsin1970, line 5: Do you avoid the title "Salām-hymn" on purpose?

DenisNosnitsin1970 (changes requested):

4001-5000/LIT4551Salam.xml
thea-m: Yes, but I am fine with reintroducing it. My first association with salām-hymn is the Synaxary salām-hymns, which are something different from the one here.

nafisa-valieva: Without arguing that my 'name is good, I referred to it as ‘Salāmta as edited in GQL’, for is appears in various forms but was edited as 'Salāmta' by the community of Lalibala.

thea-m: Thank you!
I think there is no need for a unique title, as we have the CAe reference for that. Any identifying information can of course be supplied in the abstract, or, in the case of an edition, in the bibliography. Should you have the references at hand at some point, it would be of course wonderful to insert them.
Is "Salāmtā" the modern Amharic equivalent of the Ethiopianists' "Salām-hymn"? Then I would stick for now with "salām-hymn" if "hymn" is rejected, which is more commonly, though not always clearly, used.

nafisa-valieva: Dear Dorothea, apologies for not following the rules of bm politeness! Thank you! In this case whatever choice you decide to make - whether Salām-hymn or Salām-hymn for Lalibala, I have no feelings against.

thea-m: There are no rules :) It is the question between "Salām-hymn to Lālibalā" and "Hymn to Lālibalā", no intention of hiding his name from the readers.

nafisa-valieva: Dear Dorothea, I see! But then, if Denis does not mind, I personally have no objections against 'Hymn to Lālibalā', as the incipit will reveal its salām nature :-)

DenisNosnitsin1970:
I would prefer more coherent approach to titles, esp. if they show the genre, and the titles are supplied (i.e. artificial). "Salamta" is fine, but we did not use it. Incipit can show everything, but first what I see is the title. Actually I think that salamta  is not a full equivalent to salam, and probably it is a neologism, with reference rather to the performance "giving greatings/greating" (unless the community just meant giving greetings with "Salam-hymn"). Modern church eiditons do not attend all these points. For me, "salam-hymn" is more explicit than "salamta".   But, is the recent discussion about this piece as tarafa-malk' relevant for assigning a title?  #1672

Ralph-Lee-UK:
Neither is a game changer for me - but perhaps adding Salam could be confusing as there are other hymns that start with this word, but belong to a different category, such as a Malkǝʾ-hymn.

thea-m:
@DenisNosnitsin1970

PietroLiuzzo:
When this will be looked at, may I suggest that if in the text those are linebreaks or lines of the hymn they are marked as such? with `<lb/>`, `<l></l>` or both.

thea-m:
Can this be merged @DenisNosnitsin1970 ?

PR #460, opened 2021-01-26
Gg ham

PietroLiuzzo:


nafisa-valieva:
Ok!

nafisa-valieva (commented):
I see some small errors, which we can tolerate at this point! In terms of structure, it seems to reflect our actual understanding of the dossier!

PietroLiuzzo:
in the text? it is automatically transcribed, I would be surprised if it did not have any error...

nafisa-valieva:
Thank you for your work! I think it is ready for merging!

PietroLiuzzo:
I am still making the corrections you pointed out in the word file

PR #484, opened 2021-02-23
Haleta

Ralph-Lee-UK:
Table of melodic variations for the Hallelujah አንቀጽ፡ ሃሌታ፡ for three works as sub part

### DIFF
--- a/new/LIT1990Mawase.xml
+++ b/new/LIT1990Mawase.xml
@@ -1,0 +1,316 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="work" xml:lang="en" xml:id="LIT1990Mawase">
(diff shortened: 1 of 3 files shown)


new/LIT1990Mawase.xml
```diff
@@ -0,0 +1,316 @@
...
+               <label>Antiphon of ʾAbbā Libānos</label>
+            </div>
+            <div type="textpart" corresp="LIT4498Pilgrims">
+               <label>Antiphon of Pilgrims (<foreign xml:lang="gez">ዘፈላስያን፡</foreign>)</label>
+            </div>
+            <div type="textpart" corresp="LIT4499Sunday">
+               <label>Antiphon of Sunday  (<foreign xml:lang="gez">ዘሰንበተ፡ ክርስቲያን፡</foreign>)</label>
+            </div>
+            <div type="textpart" xml:id="Haleta">
+              <label>Table of melodic variations for the Hallelujah <foreign xml:lang="gez">አንቀጽ፡ ሃሌታ፡</foreign></label>
```
DenisNosnitsin1970, line 310: As far as I can see, the spelling is አንቀጸ፡ ሃሌታ (status constructus)

DenisNosnitsin1970 (changes requested):

new/LIT1990Mawase.xml
Ralph-Lee-UK: You are quite correct - in the particular case I was looking at there is a spelling error - but it is indeed an error! I'll correct this in all three cases

Ralph-Lee-UK:
@DenisNosnitsin1970 looking at the EMIP records this also occurs with Deggwa and Soma Deggwa ms. Should we add it there as well?

DenisNosnitsin1970:
I would say, yes.

PR #486, opened 2021-02-26
Update LIT6351HymnLalibala.xml

Ralph-Lee-UK:
Added attestation in BL - as promised in our GitHub discussion


Ralph-Lee-UK:
@PietroLiuzzo - not wanting to rush you, but can you approve this edit which came from our discussion about adding an external attestation to the record?

PR #509, opened 2021-03-28
Lit5857 homily theophilus corr mk

MarcinKrawczuk:
I did some research and it turns out that the attribution given in the Ethiopian title is mistaken.


thea-m (approved):
Thank you!

PR #525, opened 2021-04-06
Lit6404 dersan4rec a

MarcinKrawczuk:
I had some doubts how to proceed here :( My intuition was to keep the record minimal and not repeat the info which is in the general record. I will send the record for recencion B, once we reach an agreement regarding the other two.


thea-m (approved):
I think that's completely fine! My experience is that readers often find it useful to have as explicit explanations as possible on the relationships of records, so I started to state and link to the general record and other recensions in `<abstract>`. But the record is completely fine as it is.

PR #539, opened 2021-05-10
Lit1594 shephe corr mk

MarcinKrawczuk:
I deleted <author>Hermas</author> because my understanding was that this filed is no longer used, but I'm not sure


thea-m (approved):
Yes, I think it's fine, thank you!

PR #553, opened 2021-05-31
Lit6454 dersan

MarcinKrawczuk:
The problem is that I have modified LIT5862OnSpiritualEaster but I missed the fact that it has already been used in ESap005 and BLorient774. So I have created this record here and I will modify the descriptions in these two ms. because I checked ESap005  and it is definitely not the same as LIT5862OnSpiritualEaster.

BLorient774 is more tricky because you cannot see it online, but in general there seems to be an almost perfect correspondence between BLorient774 and ESap005 in other cases.

Sorry for the mess. I hope this is an acceptable solution.


DariaElagina:
No worries :)

DariaElagina (approved):

PR #555, opened 2021-06-04
Update LIT1917Mashaf.xml

thea-m:


1001-2000/LIT1917Mashaf.xml
```diff
@@ -6,6 +6,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <titleStmt>
             <title xml:id="t1" xml:lang="gez">መጽሐፈ፡ አሥልጢ፡</title>
             <title corresp="#t1" type="normalized" xml:lang="gez">Maṣḥafa ʾaślǝṭi</title>
+             <title corresp="#t1" type="normalized" xml:lang="gez">Maṣḥafa ʾaśǝlṭi</title>
             <title corresp="#t1" xml:lang="en">Book of usefulness</title>
```
DariaElagina, line 10: Sorry, does the title appear then twice?

thea-m:
Probably. We could mark it up with a new ID and type="alt", but that also wouldn't be exacty correct...it's just about facilitating finding the work with whichever transliteration people search

DariaElagina:
ok, now I understand. Thank you!

DariaElagina (approved):

PR #586, opened 2021-07-20
Update LIT2051Nagara.xml

DariaElagina:


DenisNosnitsin1970 (approved):
I know this is unrelate; I just wonder where the title ገድለ፡ ማርያም፡ was picked up. It looks very unlikely here (my assumption that it refers to the Acts of another Mary).

DariaElagina:
The information on the title is derived from EAE, there are some manuscript attestations for that.

PR #593, opened 2021-08-07
Update LIT1750LahaMa.xml

MarcinKrawczuk:


1001-2000/LIT1750LahaMa.xml
```diff
@@ -31,8 +31,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             </availability>
          </publicationStmt>
          <sourceDesc>
```
DariaElagina, line 33: Here as well: do not delete these statements.

DariaElagina (changes requested):

1001-2000/LIT1750LahaMa.xml
MarcinKrawczuk: OK, done.

DariaElagina (approved):

PR #602, opened 2021-08-25
Update LIT2388Taamme.xml

DenisNosnitsin1970:
A small Miracles of the Trinity series written out.

### DIFF
--- a/2001-3000/LIT2388Taamme.xml
+++ b/2001-3000/LIT2388Taamme.xml
@@ -45,8 +45,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
          <abstract>
             <p>A general record for the Miracles of the Trinity, the first three miracles in honour of the <persName ref="PRS10775Trinity">Trinity</persName>
-            after <ref type="mss" corresp="IVorlov40"/>. The exact composition of and relations between various
-            collections of miracles dedicated to the Trinity have not been clarified yet.</p>
+            after <ref type="mss" corresp="IVorlov40"/>.</p>
          </abstract>
 
          <textClass>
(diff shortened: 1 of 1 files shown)


2001-3000/LIT2388Taamme.xml
```diff
@@ -55,22 +56,59 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                                   </ab>
+                       </div>
+                                       </div>
+                                       </div>
+                      <div type="bibliography">
          <listRelation>
             <relation name="ecrm:P129_is_about" active="LIT2388Taamme" passive="PRS10775Trinity"/>
+            <relation name="dcterms:isPartOf" active="LIT2388Taamme" passive="LIT2286ZenaSellase"/>
+              <relation name="lawd:hasAttestation" active="LIT2388Taamme" passive="IVorlov40"/>
+              <relation name="dcterms:isPartOf" active="LIT2388Taamme" passive="LIT2282Trinity" cert="medium"/>
```
thea-m, line 109: the encoding is correct, but the cert="medium" information will not be rendered and not be seen by users. Maybe it could be useful to include this information also in paragraphic form in the abstract.

thea-m (changes requested):
demons in the beer!

2001-3000/LIT2388Taamme.xml
DenisNosnitsin1970: Why not in the beer? :) I tried, but on the whole the Miracles of Trinity collections are problematic, someone should write them out, sometime.

eu-genia: Dorothea means, you should add a "desc" element to the relation saying in words that it *might* be part (or not)

eu-genia (approved):

2001-3000/LIT2388Taamme.xml
DenisNosnitsin1970: Will this work out?

2001-3000/LIT2388Taamme.xml
```diff
@@ -109,7 +108,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <relation name="dcterms:isPartOf" active="LIT2388Taamme" passive="LIT2286ZenaSellase"/>
               <relation name="lawd:hasAttestation" active="LIT2388Taamme" passive="IVorlov40"/>
               <relation name="dcterms:isPartOf" active="LIT2388Taamme" passive="LIT2282Trinity" cert="medium"/>
-         </listRelation>
+                       </listRelation>
+                       <desc>The exact composition of and relations between various collections of miracles dedicated
+                         to the Trinity have not been clarified yet.</desc>
```
eu-genia, line 113: not outside the relation though. abstract addition was fine too.

thea-m (approved):

2001-3000/LIT2388Taamme.xml
DenisNosnitsin1970: thanks!

PR #625, opened 2021-09-24
Update LIT2097OnHowt.xml

eu-genia:
Dear Marcin,
I did a small fix to the title: please check and approve if OK.
Note that you must always provide the transcription as corresp, and the English, if it corresponds to Ethiopic, should also go under the same ID (it is not t2, but t1).
Please check that your work records follow this rule.
Thank you!


PietroLiuzzo:
I am a bit puzzled by the "mark up" (the square brackets) introduced in #t1, perhaps it is just a different title?

eu-genia:
It is the same title. I took the brackets out

PietroLiuzzo (approved):

PR #637, opened 2021-11-03
Mazgaba haymanot

DariaElagina:


DenisNosnitsin1970 (changes requested):
I would suggest to quote incipit/explicit, after edition or a manuscript.

DariaElagina:
Thank you, @DenisNosnitsin1970 ! I will keep this in mind for a future pull request since I do not have access to the edition or a digitized manuscript with the work now. I only have a short excerpt transcribed by Wright, but I think that in the presence of an edition, we should rather tend to use the edition.

thea-m (approved):

PR #648, opened 2021-12-18
Lit6513 nagar

MarcinKrawczuk:


new/LIT6513Nagar.xml
```diff
@@ -57,6 +58,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <listRelation>
                     <relation name="saws:isAttributedToAuthor" active="LIT6513Nagar" passive="PRS7104Minas"/>
                 </listRelation>
```
eu-genia, line 60: add the about relation to the keyword "Masqal" (feast) and possibly also "TrueCross" (the actual cross found)

eu-genia (changes requested):

new/LIT6513Nagar.xml
MarcinKrawczuk: Can you please explain, keywords, do you mean this: https://betamasaheft.eu/authority-files/TrueCross/main? But there's no record number here.

eu-genia: You do not need the record number, simply the keyword (passive="Masqal", passive="TrueCross")

eu-genia (approved):

PR #679, opened 2022-01-28
Create LIT6553SacerPrayers.xml

DenisNosnitsin1970:
New record

### DIFF
--- a/new/LIT6553SacerPrayers.xml
+++ b/new/LIT6553SacerPrayers.xml
@@ -6,6 +6,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <fileDesc>
             <titleStmt>
                 <title xml:id="t1" type="main" xml:lang="en">Sacerdotal prayers for the Divine Office</title>
+                <title xml:id="t2" xml:lang="en">Daily prayers</title>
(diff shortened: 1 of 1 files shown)


thea-m (changes requested):
Could you maybe add the bibliographic reference from which the designation is taken (Velat, I bleieve)?

eu-genia:
Agree, maybe adding to the abstract "referred to as priéres sacerdotales in `<bibl><ptr target="bm:Velat1966MeerafTrad"/><citedRange>161</citedRange></bibl>`" could be helpful to understand the term; one could also add other terms if such were used somewhere (how does Habtemichael call them, if he groups them at all?)

DenisNosnitsin1970:
The appearance of Velat in the discussion above is accidental. The major issue is the way of  classifying the content of "Mashafa qeddase" into three main parts (Daily prayers; Ser'ata qeddase; Anaphoras). I followed the practice of Getatchew Haile who grouped the mentioned items together; he called them like this, but only in the EMML I (I can certainly mention him in the bibliography; later in the same volume and futher he changed to somewhat ambiguous but economic "Office prayers"). Some catalogues do group the items together (Wright) using various designations (I would prefer that way), some treat them separately (VOHD, four parts, or others).

DenisNosnitsin1970:
Well, formally the author of EMML I was Macomber :)

PR #690, opened 2022-02-10
Solomon2022

sologebre:


new/LIT6557Mashafat.xml
```diff
@@ -0,0 +1,55 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6557Mashafat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Maṣḥafa tārik</title>
```
thea-m, line 8: Could you also add the title in fidal?

new/LIT6557Mashafat.xml
```diff
@@ -0,0 +1,55 @@
...
+                <language ident="gez">Gǝʿǝz</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="SG" when="2022-02-07">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography"/><!---->
```
thea-m, line 52: It would be good to add at least the incipit and explicit, and possibly an abstract briefly summarising the content. And in the case of this lengthy work, maybe some subdivision into the major parts (if they can be identified) might also be useful. The normal users of BM will not have access to the manuscript description for a long time, so any information that disambiguates this work will be very useful.

thea-m (changes requested):
Thank you, Solomon! Technically everything is fine, but it would be extremely useful if you could possibly elaborate this record a little bit.

new/LIT6557Mashafat.xml
sologebre: Thank you, Dorothea!

new/LIT6557Mashafat.xml
```diff
@@ -49,7 +54,30 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     </teiHeader>
     <text>
         <body>
-            <div type="bibliography"/><!---->
+<div type="edition">
```
thea-m, line 57: I think that there are a few small problems here with the nesting of textparts. Do I understand it correctly, that the text is divided in the parts "Genealogyofkings", "GenealogyofMary", "GenealogyofEthiopiankings", and "LIT6568YanagastatKebr", and that you give the incipit for each of these textparts? If so, each of these textparts should be their own  `<div type="textpart">` and then contain a further  `<div type="textpart" subtype="incipit">` in which the incipit should be encoded. But maybe I misunderstood.

thea-m (changes requested):
Thank you! Sorry for the late reaction. These seem really interesting!

eu-genia:
it would be nice if work records when created are also merged for everyone to know that they are there, could you please finalize the PR? thank you!

thea-m:
I will continue to work on this.

eu-genia:
@thea-m can you maybe fix the conflicts and merge this? thank you

thea-m:
If it's alright, I will deal with this when I come back in December :)

thea-m:
https://github.com/BetaMasaheft/Works/pull/1342

PR #697, opened 2022-02-18
Works in bdlaethg33

MassimoVilla:
Here come three new work records from MS g.33


thea-m (changes requested):
Thank you! Would it be possible to add the explicits as well? I know it takes time, but I find them really useful for identifying texts, and people will have to wait a long time for the manuscript descriptions.

PR #719, opened 2022-03-07
deleted LIT2860RepCh141 and LIT5777MalkKiros , doublets of  LIT2851RepCh133

thea-m:
(https://github.com/BetaMasaheft/Documentation/issues/1984)


2001-3000/LIT2851RepCh133.xml
```diff
@@ -4,7 +4,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title xml:id="t1">Malkǝʾ-hymn to St Qirqos</title>
+            <title xml:id="t1" xml:lang="en">Malkǝʾ-hymn to St Qirqos</title>
+            <title xml:lang="en" xml:id="t2">Malkǝʾ-hymn to St Kiros</title>
```
DenisNosnitsin1970, line 8: Actually, this is very strange, Qirqos and Kiros cannot be the same.

DenisNosnitsin1970 (changes requested):

2001-3000/LIT2851RepCh133.xml
thea-m: Only Kiros is mentioned in the manuscripts, "cyr" or "cyrus" in the catalogues, of which I think that Qirqos was our erroneous translation. I think Qirqos can be deleted.

eu-genia (approved):

PR #720, opened 2022-03-07
Update LIT2870RepCh151.xml

thea-m:


2001-3000/LIT2870RepCh151.xml
```diff
@@ -36,7 +36,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
          <abstract>
            <p>The attribution of the hymn to John the Evangelist is based on <bibl><ptr target="bm:Chaine1913Rep"></ptr></bibl>. However,
-           according to <bibl><ptr target="bm:Wright1877BM"></ptr></bibl> it is addressed do John the Baptist, whereas
+           according to <bibl><ptr target="bm:Wright1877BM"></ptr></bibl> it is addressed to John the Baptist, whereas
```
DenisNosnitsin1970, line 39: I think the re-identification will remain shaky untill we extend a little bit the knowledge of the text body; here we cannot see more those six words that we have in Chaine's publication. Can we get and more text, from one source or another, especially in disambiguation cases?

DenisNosnitsin1970 (changes requested):

2001-3000/LIT2870RepCh151.xml
DenisNosnitsin1970: Somehow it is difficult to digest this gaps - John the Baptis - the Evangelist - King Yohannes. Probably the author knew the subject he was writing about.

eu-genia (approved):

PR #726, opened 2022-03-12
Fixing div edition nesting

eu-genia:
While going through the records I noticed that sometimes `div="edition"` is nested within `div="bibliography"` under `body`; I have fixed all such cases, it would be great if we all in the future try to avoid this, `div="edition"` should be the next level after `body`. Thank you so much!!!

xpath test that edition is first-level under body `$config:collection-rootW//t:body/t:div[t:div[@type='edition']]`
/second part to https://github.com/BetaMasaheft/Works/pull/725/


thea-m:
Thank you, I'll keep it in mind.

PR #724, opened 2022-03-10
deleted LIT6434Hymn, doublet of LIT2981RepCh262

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/1993

PR #731, opened 2022-03-15
deleted LIT6434Hymn, doublet of LIT2981RepCh262

thea-m:


thea-m:
https://github.com/BetaMasaheft/Works/pull/724

thea-m:
Here, there seem to be no conflict, but please check and approve so that we are sure.

PR #739, opened 2022-03-15
Lit6523 feleksinos

MarcinKrawczuk:


thea-m (approved):

eu-genia:
Note that the record has div edition inside of div bibliography, am making a pull request with a corrected version

thea-m:
Now I see it 🙈

PR #740, opened 2022-03-15
Lit6522 petros

MarcinKrawczuk:
It really is ስኒዶስ in the manuscript, it's not a typo


thea-m (approved):
Thanks a lot! It would be really helpful if in the future you could open a text identification issue before creating a new record (they are quite important for our documentation) - sorry if there was one and I missed it!

PR #752, opened 2022-03-21
LIT 3143 3972 4336

DenisNosnitsin1970:
I tried to complete and disambiguate the genealogies related to Takla Hazmanot.

### DIFF
--- a/3001-4000/LIT3143Ledata.xml
+++ b/3001-4000/LIT3143Ledata.xml
@@ -48,7 +48,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </textClass>
          <abstract>
              <p>The monastic genealogy of <persName ref="PRS9162TaklaHa">St Takla Hāymānot</persName> that shows
-             his spiritual descendance from <persName ref="PRS1385Adam">St Anthony</persName> and tells about his disciples
+             his spiritual descendance from <persName ref="PRS1962Anthony">St Anthony</persName> and tells about his disciples
              and their spiritual successors. In part based on <ref type="work" corresp="LIT3973GadlaT">the Vita of Takla Hāymānot</ref>,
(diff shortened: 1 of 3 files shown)


thea-m (changes requested):
Thank you! Everything that is there seems fine to me, and is very helpful. But I would be grateful if you could include an abstract in each record which very briefly describes the work in question and , if relevant, its relation/disambiguation from the other works. I think that this is one of the cases where paragraphic information is appropriate.

DenisNosnitsin1970:
I have tried.

thea-m:
Here, in this PR? I don't see an abstract in the records.

DenisNosnitsin1970:
Now. No idea why it got stuck in the pipe.

3001-4000/LIT3143Ledata.xml
```diff
@@ -39,20 +35,62 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
             <keywords scheme="#ethioauthlist">
                <term key="HistoryAndHistoriography"/>
+                <term key="ChristianLiterature"/>
+                <term key="Hagiography"/>
             </keywords>
+
          </textClass>
+         <abstract>
+             <p>The monastic genealogy of <persName ref="PRS9162TaklaHa">St Takla Hāymānot</persName> that shows
+             his spiritual descendance from <persName ref="PRS1385Adam">St Anthony</persName> and tells about his disciples
```
thea-m, line 51: Adam or Anthony?

thea-m (changes requested):
Thank you! One minor question.

3001-4000/LIT3143Ledata.xml
DenisNosnitsin1970: Anthony.

thea-m (approved):

PR #762, opened 2022-03-23
Lit6585 rh pentecost

MarcinKrawczuk:
https://github.com/BetaMasaheft/Documentation/issues/2018


new/LIT6583RHMaryam.xml
```diff
@@ -0,0 +1,60 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
```
thea-m, line 1: Did some branch confusion happen here? LIT6583RHMaryam was created through https://github.com/BetaMasaheft/Works/pull/764/files as well. If this file here is a redundant doublet, I think the best would be if you deleted it from this branch.

thea-m (changes requested):
Logistical question :) The other record is fine. Just notice the quick correction I carried out with the wrong nesting of the bibliography and edition. This happens in some records, as Zhenia pointed out (we should have noticed before, our mistake)

new/LIT6583RHMaryam.xml
MarcinKrawczuk: Sorry, as usual I don't know how to sort these branch problems... Do you mean that I should switch to branch LIT6585RHPentecost and then delete the file LIT6583RHMaryam from my GitHub folder?

thea-m: Yes, exactly, that was the problem. I did it now quickly, sorry this got lost among other things for so long.

thea-m (approved):

PR #778, opened 2022-03-28
Create LIT6606JohnAlms.xml

MarcinKrawczuk:
https://github.com/BetaMasaheft/Documentation/issues/2018


thea-m (approved):

thea-m (changes requested):
Actually, now I noticed some ID confusion with this record and LIT6607JohnFast - the IDs and the titles seem to be interchanged? 
In theory, the alphabetical part of the IDs is irrelevant, but since we cannot help but read it and also use it for orientation, maybe you could fix this? I think the easiest would be - if my interpretation is right - to copy paste and exchange the complete content of both records. This would allow to keep the IDs. If you prefer to change the IDs, it would be necessary to merge the two pull requests and then make a new one for the deletion of the files, and again a new one for the creation of the new records.
I hope you understand what I mean :) My writing seems a bit confused to myself.

thea-m (approved):

MarcinKrawczuk:
Thank you very much for pointing that out. However I won't be able to work
on it this week, because I'm going away for the Easter period. But I will
fix it as soon as possible.



pon., 11 kwi 2022 o 14:00 thea-m ***@***.***> napisał(a):

PR #787, opened 2022-04-24
Lit6611 sal yoh mat

MarcinKrawczuk:


thea-m (changes requested):
LIT6611SalYohMat is fine, but I think there is some branch confusion? :) Since LIT6613DerBetYoh has already been merged, you could delete the record from this branch and complete LIT6612 and LIT6614. We could close the PR for now, and you could open it again when that is done. Would that work for you?

MarcinKrawczuk:
Recently it happens all the time, I still find it so confusing working with
the branches :( I would like to keep it 1 work record=1 one branch (the
same name).  I'll try to fix it when I'm back home from the classes.

pon., 25 kwi 2022 o 09:11 thea-m ***@***.***> napisał(a):

thea-m:
I'm really sorry, for me it was the same when we implemented this. With more practice it will become somewhat automatic. If you have any idea how the process could be made easier, please let me know!

MarcinKrawczuk:
What does it mean "delete the record from this branch and complete LIT6612 and LIT6614". How do I delete a record from a branch?

thea-m:
You can delete by making sure you are on this branch and then deleting the file LIT6613DerBetYoh either directly in atom or in your project folder, like you would delete any file. If you then commit the deletion to GitHub, the conflict here should disappear.

MarcinKrawczuk:
OK, done. I hope it brings us closer to the solution.

wt., 26 kwi 2022 o 11:05 thea-m ***@***.***> napisał(a):

thea-m:
Excellent, thank you!

thea-m (approved):

PR #795, opened 2022-04-25
Update LIT1286Dersan.xml

MarcinKrawczuk:
A little but hopefully useful update


eu-genia (approved):
thank you, I only updated the title statement on the occasion

PR #845, opened 2022-06-30
correction

MarcinKrawczuk:
So if it's ok and you merge it, I will correct the ms records


eu-genia (changes requested):
You forgot to change the xml:id of the file inside; it must be the same as the filename. Thank you.

eu-genia (approved):

PR #861, opened 2022-08-29
Lit2053 nagara corr mk

MarcinKrawczuk:
I'm submitting Nagara Sanbat for review. It would be nice to create a stub for the ms from which I've taken the incipit but that's beyond my competence. I hope you like it :)

### DIFF
--- a/2001-3000/LIT2053Nagara.xml
+++ b/2001-3000/LIT2053Nagara.xml
@@ -36,7 +36,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
-         <abstract><p>Nagara sanbat is a short work of the Beta ʾƎsraʾl. </p></abstract>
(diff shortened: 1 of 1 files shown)


2001-3000/LIT2053Nagara.xml
```diff
@@ -49,11 +51,25 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
          <change when="2017-09-05" who="DR">Corrected title</change>
+         <change who="MKr" when="2022-08-29">Added summary and incipit based on the info provided by Sophia Dege-Müller</change>
       </revisionDesc>
    </teiHeader>
+   <text>
+        <body>
+          <div type="bibliography"/>
+
+          <div xml:lang="gez" type="edition">
+            <note>The incipit is taken Faitlovitch MS 13</note>
```
SophiaD-M, line 62: Maybe add "from" - the incipit is taken "from" Faitlovitch...
what do you think?

SophiaD-M: Oh and also from which folio is the incipit?

SophiaD-M (approved):

MarcinKrawczuk:
OK, corrected both!

wt., 30 sie 2022 o 14:24 SophiaD-M ***@***.***> napisał(a):

PR #863, opened 2022-08-31
🐎 ci: update PR validation

duncdrum:
!!WARNING!! For every merge we need to monitor the production server so it has time to complete the gitsync without crashing.

see BetaMasaheft/Documentation#2192


line-o (approved):
🤞🏼

PR #877, opened 2022-09-20
new records

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2184

### DIFF
--- a/new/LIT6715PPQwerdat.xml
+++ b/new/LIT6715PPQwerdat.xml
@@ -1,0 +1,70 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6715PPQwerdat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against <foreign xml:lang="gez">qʷərḍat</foreign>, containing <foreign xml:lang="gez">ʾasmāt</foreign></title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DR"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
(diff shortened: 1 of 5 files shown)


new/LIT6715PPQwerdat.xml
```diff
@@ -0,0 +1,70 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6715PPQwerdat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against <foreign xml:lang="gez">qʷərḍat</foreign>, containing <foreign xml:lang="gez">ʾasmāt</foreign></title>
```
MassimoVilla, line 8: I am just wondering whether we may use - for the benefit of the reader - the English term in the title, perhaps with the Ethiopic term in brackets, e.g. `Protective prayer against stomachache (<foreign xml:lang="gez">qʷərḍat</foreign>), containing ...`

new/LIT6718PPAqweyasat.xml
```diff
@@ -0,0 +1,78 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6718PPAqweyasat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against <foreign xml:lang="gez">ʾaqʷəyāṣāt</foreign> (shinbone?) and magicians</title>
```
MassimoVilla, line 8: Unfortunately  I cannot figure out the implications of ጸሎተ፡ አቍያጻት and why shinbones are linked to magicians

MassimoVilla (changes requested):
Thank you Thea. Just very few remarks. Magic prayers are always tricky in my view.

thea-m:
Thank you, @MassimoVilla ! It's always tricky with these prayers. I'm not completely confident with the identification of the illnesses with modern English terms, so I used the Geez terms. But you are right, for those unfamiliar with these texts that would be confusing. @karljonaskarlsson any opinion?

karljonaskarlsson:
I think it's preferable to use English terms as far as possible, maybe with question marks, as Massimo suggested. Using English may open up for mistakes (I would rely completely on dictionaries and other sources), but it also adds value for the average reader, so I think it's worth it.

karljonaskarlsson:
English + Ge'z, I mean

MassimoVilla (approved):

PR #878, opened 2022-09-20
Update LIT6719prologue.xml

karljonaskarlsson:
I created the work record (with Zhenia's help for the first steps).

### DIFF
--- a/new/LIT6719prologue.xml
+++ b/new/LIT6719prologue.xml
@@ -5,7 +5,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">prologue</title>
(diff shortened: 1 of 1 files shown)


thea-m (approved):
Everything seems fine to me!

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2236
It is always better to link to the work creation issue for documentation

SophiaD-M:
May I suggest to start with capital P - LIT6719Prologue ?
I somehow think that this is usually the case? Or am I wrong?

thea-m:
Yes, it's the convention, but there is no problem if it's not respected in one case. Better to leave it as created and not change an existing ID

SophiaD-M:
but was it not created just now? Would this not be the best possible time to change the ID?

new/LIT6719prologue.xml
```diff
@@ -30,7 +32,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p/>
+                <p>A prologue to the Book of Psalms ascribed to Eusebius of Caesarea. It usually consists of an introduction,
+                    a list of the authors of psalms (frequently with the number of psalms ascribed to each of them
+                    appearing next to the name), and an section regarding the naming of the Book of Psalms, followed
```
MassimoVilla, line 37: a section

PR #880, opened 2022-09-20
Update LIT6719prologue.xml

karljonaskarlsson:


thea-m:
You also have to rename the file, so that the xml:id and name of file are identical

thea-m:
In oxygen (or the folder on your computer in which you keep the GitHub files). Like any file renaming, right mouse click etc.

thea-m (approved):

PR #895, opened 2022-10-11
BDLaethg43

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2186

### DIFF
--- a/new/LIT6751PPAynat.xml
+++ b/new/LIT6751PPAynat.xml
@@ -1,0 +1,80 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6751PPAynat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against the evil eye (<foreign xml:lang="gez">ʿāynat</foreign>),
+                    containing <foreign xml:lang="gez">ʾasmāt</foreign></title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DR"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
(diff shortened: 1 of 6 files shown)


new/LIT6751PPAynat.xml
```diff
@@ -0,0 +1,80 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6751PPAynat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against the evil eye (<foreign xml:lang="gez">ʿāynat</foreign>),
```
MassimoVilla, line 8: Etymologically ʿāynat is ʿaynat (Cp, Leslau, Comp. Dict. p. 80a), unless a different spelling form has been established and adopted in BM

new/LIT6754PPQwertmat.xml
```diff
@@ -0,0 +1,71 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6754PPQwertmat" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Protective prayer against rheumatism (?) (<foreign xml:lang="gez">qʷǝrṭmāt</foreign>),
```
MassimoVilla, line 8: One ǝ is missing in qʷǝrṭmāt (three consecutive consonants are not allowed)

MassimoVilla (changes requested):
Thank you!

new/LIT6751PPAynat.xml
thea-m: You're right (and it was never discussed), but in my estimation the spelling with ዓ is the pervasive one, so I would tendentially be against a normalisation in this case.
Cp. https://betamasaheft.eu/as.html?query=%E1%8B%93%E1%8B%AD%E1%8A%90%E1%89%B5&AttestedInType=1&xmlid= and https://betamasaheft.eu/as.html?query=%E1%8B%90%E1%8B%AD%E1%8A%90%E1%89%B5&AttestedInType=1&xmlid=

new/LIT6754PPQwertmat.xml
thea-m: True. Strelcyn has qʷǝrṭǝmāt, so let's follow him?

MassimoVilla: Let's follow Strelcyn!

new/LIT6751PPAynat.xml
MassimoVilla: ዓይነት፡ is indeed more frequent, even though the etymological spelling form remains ዐይነት. However, let's stick to the popular form. We might discuss and standardize these forms at a later stage

MassimoVilla (approved):

new/LIT6751PPAynat.xml
karljonaskarlsson: It's an interesting discussion! I believe that there might be a diachronic dimension to many of these cases as well, but it remains to be studied more in depth.

PR #908, opened 2022-10-20
Lit6755 magas

MarcinKrawczuk:
https://github.com/BetaMasaheft/Documentation/issues/2252

Could you please check especially whether the Arabic is correct, I don't feel very confident in it.


thea-m (approved):
everything fine, I just added the transcription

PR #917, opened 2022-11-16
Further text in ms e.39

MassimoVilla:
Here comes the last new text in MS BDLaethe39


thea-m (changes requested):
It was attested in one manuscript as part of a Sawāsəw, right? If that is true, a relation could be added here. If not, everything is fine, thanks a lot!

thea-m (approved):

PR #920, opened 2022-11-21
Update LIT5921SalamLaki.xml

MarcinKrawczuk:
I think the relation needs to be corrected, otherwise it doesn't display correctly in the app


thea-m (approved):
Thank you!

PR #891, opened 2022-10-07
new records BDLaethg40

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2212

PR #922, opened 2022-11-23
Dr 22 11 corrections

thea-m:
https://github.com/BetaMasaheft/Works/pull/891

### DIFF
--- a/new/LIT6740PPWegat.xml
+++ b/new/LIT6740PPWegat.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="en" xml:id="t1">Protective prayer in Gǝʿǝz and Amharic against stomachache (<foreign xml:lang="gez">wəgʿāt</foreign>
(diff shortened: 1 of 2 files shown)


new/LIT6740PPWegat.xml
```diff
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="en" xml:id="t1">Protective prayer in Gǝʿǝz and Amharic against stomachache (<foreign xml:lang="gez">wəgʿāt</foreign>
+                <title xml:lang="en" xml:id="t1">Protective prayer in Gǝʿǝz and Amharic against stomachache (<foreign xml:lang="gez">wəgāt</foreign>
```
MassimoVilla, line 8: Sorry, is it wəgāt the proper form? I think I missed something. Dillmann gives ውግአት (https://betamasaheft.eu/Dillmann/lemma/L34927d16d21b476a8074e6637e1a0a2e). Perhaps wəgāt  is Amharic

thea-m: you're right, it's Amharic (but not so rare in scrolls, I think...not sure though). But since this is the form attested here, I would not like normalising it in the title. Should we just change the language here to Amharic?

MassimoVilla: I would change the language to Amharic then!

MassimoVilla (approved):

PR #937, opened 2022-12-19
Create LIT6801KwellomuMalaekt.xml

DenisNosnitsin1970:
New record. https://github.com/BetaMasaheft/Documentation/issues/2289

### DIFF
--- a/new/LIT6801KwellomuMalaekt.xml
+++ b/new/LIT6801KwellomuMalaekt.xml
@@ -1,0 +1,78 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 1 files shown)


new/LIT6801KwellomuMalaekt.xml
```diff
@@ -0,0 +1,78 @@
...
+            <change who="DN" when="2022-12-19">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT6801KwellomuMalaekt" passive="SinaiEt004"/>
+                    <relation name="lawd:hasAttestation" active="LIT6801KwellomuMalaekt" passive="BAVet15"/>
+                    <relation name="saws:isDifferentTo" active="LIT4965Kwellomu" passive="BAVet15"/>
```
thea-m, line 59: I find the relations a bit confusing. The work is attested in two manuscripts, but is different from one of them? saws:isDifferentTo should not be used between a manuscript and a work record - can you find a better way to express the situation, possibly using `<desc>` inside the relation or describing the state in `<abstract>`?

thea-m (changes requested):

new/LIT6801KwellomuMalaekt.xml
eu-genia: I think it should have been
`<relation name="saws:isDifferentTo" active="LIT6801KwellomuMalaekt" passive="LIT4965Kwellomu"/>`

thea-m (approved):

PR #958, opened 2023-03-17
added new textparts

MassimoVilla:
Dear all, here comes a PR with few updates about the textparts of the Maṣḥafa gǝnzat


thea-m (approved):
Thank you! If in the course of the manuscript description you'll also copy the corresponding explicits, it could be useful to include them here as well.

MassimoVilla:
Great thank you. I'll provide the explicits as well then!

thea-m (approved):

PR #961, opened 2023-03-28
deleted LIT2212Questi, doublet of LIT5193HistoryBenjamin

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2327
Thank you!
Do you know spontaneously if they are questions of, or to, Theodore? If both titles are used, we could add one with "to" as well.


CarstenHoffmannMarburg (approved):
I have no knowledge on that.

MarcinKrawczuk:
EAe uses "Questions **of** Theodore", the same seems to be used for the Coptic version so I would stick with "of".

MarcinKrawczuk (approved):

PR #962, opened 2023-03-30
clarification

thea-m:


DenisNosnitsin1970:
That is fine, but I would say clarification is not sufficient in order to distinguish between the versions. Will someone add a bit of incipits?

thea-m:
Not me right now :) The editions of each version are given.

PR #964, opened 2023-04-17
restoring CAe 3495

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/2326


thea-m:
Thank you! All is fine. Do you think it would be necessary to add a short abstract for those who do not read the relations? Since there must have been confusion surrounding the file, as it had been deleted.

eu-genia:
I have added the abstract but I assume we shall be restoring most of the IDs as they have independent authorships / circulations outside of Ethiopia

thea-m (approved):

PR #983, opened 2023-05-10
Update LIT6459AstronomicalTreatise.xml

DariaElagina:


6001-7000/LIT6459AstronomicalTreatise.xml
```diff
@@ -30,11 +36,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
             <abstract>
-                <p/>
+                <p>A cosmological treatise of a European provenance, possibly translated from a Latin textbook on astronomy by a Jesuit missionary in 
+                    Ethiopia in <date>the first half of the 17th century</date> (<bibl>
+                        <ptr target="bm:HoffmannGeus2020Cosmology"></ptr>
+                        <citedRange unit="page">62</citedRange>
+                    </bibl>). The text was believed to be transmitted only in one 
+                    manuscript (<ref type="mss" corresp="BNFabb109"></ref>) employed in all three existing editions, 
+                    however, at least one further witness could be identified (<ref type="mss" corresp="BLorient816"></ref>). A preliminary 
+                comparison of the witnesses show some deivergencies in readings.</p>
```
CarstenHoffmannMarburg, line 46: divergencies

CarstenHoffmannMarburg (changes requested):

6001-7000/LIT6459AstronomicalTreatise.xml
CarstenHoffmannMarburg: Thank you for addressing this text!

6001-7000/LIT6459AstronomicalTreatise.xml
```diff
@@ -30,11 +36,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
             <creation/>
             <abstract>
-                <p/>
+                <p>A cosmological treatise of a European provenance, possibly translated from a Latin textbook on astronomy by a Jesuit missionary in 
+                    Ethiopia in <date>the first half of the 17th century</date> (<bibl>
+                        <ptr target="bm:HoffmannGeus2020Cosmology"></ptr>
+                        <citedRange unit="page">62</citedRange>
+                    </bibl>). The text was believed to be transmitted only in one 
+                    manuscript (<ref type="mss" corresp="BNFabb109"></ref>) employed in all three existing editions, 
+                    however, at least one further witness could be identified (<ref type="mss" corresp="BLorient816"></ref>). A preliminary
```
thea-m, line 45: was the additional witness identified by bm:HoffmannGeus2020Cosmology or by you now? (Good general practice to try to avoid passive when pointing to the orginators' of research findings, I think)

thea-m (changes requested):
Thank you :)

6001-7000/LIT6459AstronomicalTreatise.xml
DariaElagina: It was me (unless I have overseen something), but I do not know how to state this in such a contribution. It is equally strange to use the first person (because the contribution might be reworked by other authors in the future) and to write down my own name as if I were speaking about somebody else.  Do you have any solution?

thea-m: I understand, but in my opinion a bit of strangeness would be worth the clarity. I would go for full name: "however, Daria Elagina identified a further witness..."

thea-m (approved):

thea-m (approved):

PR #987, opened 2023-05-16
deleted LIT1663Homily, doublet of LIT1183Baenta

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2351

I think that LIT1663Homily can be safely deleted. I've kept LIT1183Baenta for the moment and added a reference to it in the Hāymānota ʾabaw record.
Please also check the English title - I think that it's useful to include one (if only to have the record among the results if someone searches for "homily") and went with "life-giving" instead of "salvatory", but it can be changed.


AaronButts:
Yes, 'life-giving' is better for me in English. 

I agree that it is fine to delete LIT1663Homily.

PR #994, opened 2023-06-01
Create LIT6833Salam

CarstenHoffmannMarburg:
I created an ID for LIT6833Salam, that is to be found in BMLacq680 and that was discussed in Issue #2367.

### DIFF
--- a/new/LIT6833Salam.xml
+++ b/new/LIT6833Salam.xml
@@ -1,0 +1,66 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6833Salam" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Salām lakkəmu Ṣādəqān wa-Samāʿət</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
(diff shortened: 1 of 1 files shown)


new/LIT6833Salam.xml
```diff
@@ -0,0 +1,66 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6833Salam" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Salām lakkəmu Ṣādəqān wa-Samāʿət</title>
```
thea-m, line 8: samāʿt
I also would spell all words after the first small since they are not really proper names - but I'm not sure

new/LIT6833Salam.xml
```diff
@@ -0,0 +1,66 @@
...
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
+                    <term key="Prayers"/>
```
thea-m, line 38: Poetry?

thea-m (changes requested):
Very nice, thank you!

new/LIT6833Salam.xml
CarstenHoffmannMarburg: I was wondering, if I delete the ə in samāʿ(ə)t, I need also delete the ə in ṣād(ə)qān, not? If it is no disturbance to anyone, I can be more archaic and less traditional in this.
No, they are not proper names. So, if this is the rule, I can write all translitterations in uniform small letters.

new/LIT6833Salam.xml
CarstenHoffmannMarburg: Done. I think, we can count every Salam as poetry by definition, even though I do not find any rhyme in it.

thea-m: it rhymes on -t and in the second part on -s :) samāʿt, ʾəlat, təʾgəśt...

CarstenHoffmannMarburg: Very good! You see, I am not so much in rhyme in poetry to discover that!

new/LIT6833Salam.xml
thea-m: I think that both shwas are not etymologically equivalent - samāʿt is the plural pattern qatalt, and the a is only long due to the laryngeal, while in ṣādəqān the shwa is part of the pattern, and I'm not aware of a vowel reduction. But I'm open to other interpretations!

(There is this list based on decisions taken in TraCES, but it's not really updated regularly: https://betamasaheft.eu/Guidelines/?id=dubious-spelling)

CarstenHoffmannMarburg: So, I am convinced to spell ṣādəqān, but samāʿt.

CarstenHoffmannMarburg: It might be useful to have a longer list in the guidelines, for instance for lakkəmu vs. lakəmu etc.

thea-m (approved):

new/LIT6833Salam.xml
thea-m: That's true. The basic assumption is that the TraCES system is used, but there are some differences, and updates might always be necessary. Feel free to edit the guidelines! Alessandro could then also be assigned in a guidelines PR

PR #1003, opened 2023-06-16
Created LIT for Mawāśəʿt za-ʾArsimā

CarstenHoffmannMarburg:
I created a LIT ID for Mawāśəʿt za-ʾArsimā, as attested in BMLor148 and discussed in Issue [#2381](https://github.com/BetaMasaheft/Documentation/issues/2381). I added this ID also to the general record as `<div>`, because quiet a good number of Antiphons is listed there as `<div>` already.

### DIFF
--- a/1001-2000/LIT1990Mawase.xml
+++ b/1001-2000/LIT1990Mawase.xml
@@ -74,242 +74,242 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             </listBibl>
          </div>
          <div type="edition">
-            <div type="textpart" corresp="LIT4424JohnTheBaptist">
+            <div type="textpart" corresp="LIT4424JohnTheBaptist" xml:id="JohnB">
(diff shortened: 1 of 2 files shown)


1001-2000/LIT1990Mawase.xml
```diff
@@ -309,7 +309,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <div type="textpart" xml:id="Haleta">
               <label>Table of melodic variations for the Hallelujah <foreign xml:lang="gez">አንቀጸ፡ ሃሌታ፡</foreign></label>
             </div>
-
+            <div type="textpart" xml:id="LIT6841MawArsima">
```
eu-genia, line 312: please take a look at the other textparts
`xml:id `is internal ID, custom, can by any string, but it may not start with `LIT`
and then if the textpart has been upgraded to an independent CAe entry because of independent circulation then you also can have a `corresp `which points to that record (in this case `LIT6841MawArsima`)

eu-genia (changes requested):

eu-genia:
when a PR is linked to an issue then can you please insert in the first message the link to the issue, not the issue number? (I have been correcting all but it would be nice if I do not have to)

eu-genia (approved):

PR #1006, opened 2023-06-22
Work in BDLeu6

MassimoVilla:
Dear both,
here comes the record for the monastic genealogy in MS BDLeu6

### DIFF
--- a/new/LIT6842TentaLe.xml
+++ b/new/LIT6842TentaLe.xml
@@ -1,0 +1,73 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6842TentaLe" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ጥንተ፡ ልደቶሙ፡ ለመነኮሳት፡</title>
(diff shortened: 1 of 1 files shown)


new/LIT6842TentaLe.xml
```diff
@@ -0,0 +1,73 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6842TentaLe" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ጥንተ፡ ልደቶሙ፡ ለመነኮሳት፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Ṭǝnta lǝdatomu lamanakosat</title>
```
thea-m, line 9: I think it would be good for quick disambiguition if Dabra Dāmmo would be mentioned in the title

new/LIT6842TentaLe.xml
```diff
@@ -0,0 +1,73 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
```
thea-m, line 35: Could you add a short abstract where the difference from the other monastic genealogies and the relation to Dabra Dāmmo is briefly stated? It might also be useful to add relations below to state the difference from the other records

thea-m (changes requested):
Thank you! I hope that my suggestions are not too much work.

new/LIT6842TentaLe.xml
MassimoVilla: Not much trouble! This genealogy seems interesting for the history of Dabra Dāmmo, perhaps it might be useful to publish it in a short contribution as an outcome of our project.

thea-m (approved):
Thank you!!

PR #1017, opened 2023-07-12
Update LIT4713Liton.xml

CarstenHoffmannMarburg:
I wish to encode new Mawāśəʿt as Leonhard Bahr has suggested and we are discussing in Issue #2398. For the beginning, I start with an update to LIT4713Liton, as proposed by Leonard with adding titles, an explicit and an edition in the bibliography.


thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2398

CarstenHoffmannMarburg:
Can this update be approved and merge or are there any checks or modifications still necessary?

PR #1034, opened 2023-11-29
schwa fixing in works

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/2443


thea-m (approved):

eu-genia:
dear @thea-m please do not merge big commits all at once, it may cause a system collapse. for 22 files it takes up to 2 hours for the merge to go through. so only one by one, every now and then, this is why they have not yet been merged. most probably the changes will not make it to the app. thank you!

PR #1052, opened 2024-02-05
Update records on biblical books

CarstenHoffmannMarburg:
I updated the records on the book of Daniel according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2478#issuecomment-1924219354 and added a minor bibliographical note to LIT1340Enoch.

### DIFF
--- a/1001-2000/LIT1259Daniel.xml
+++ b/1001-2000/LIT1259Daniel.xml
@@ -35,8 +35,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <abstract>
-            <p>General ID for the biblical <ref type="work" corresp="wd:Q80115">Book of Daniel</ref>, referring to a recension that does 
-               not include the stories of <ref type="work" corresp="LIT3194Susanna">Susanna</ref>, 
+            <p>General ID for the biblical Book of Daniel, referring to a recension that does not include the stories of 
+               <ref type="work" corresp="LIT3194Susanna">Susanna</ref>, 
                <ref type="work" corresp="LIT3195Bel">Bel and the Dragon</ref>, 
                the <ref type="work" corresp="LIT2250salota">Prayer of Azariah and the Three Holy Children</ref>, which belong to the 
                so-called deuterocanonical books, or to a text for which this is unknown.</p>
@@ -53,7 +53,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <body>
          <div type="bibliography">
             <listRelation>
-               <relation name="saws:formsPartOf" active="LIT1259Daniel" passive="LIT3529Daniel"/>
+               <relation name="ecrm:CLP46i_may_form_part_of" active="LIT1259Daniel" passive="LIT3529Daniel"/>
             </listRelation>     
          </div>
       </body>
--- a/1001-2000/LIT1259Daniel.xml
+++ b/1001-2000/LIT1259Daniel.xml
@@ -35,9 +35,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <abstract>
-            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the 
-               Dragon, the Prayer of Azariah and the Three Holy Children, which belong to the so-called deuterocanonical books, or to a or to a 
-               text for which this is unknown.</p>
+            <p>General ID for the biblical <ref type="work" corresp="wd:Q80115">Book of Daniel</ref>, referring to a recension that does 
+               not include the stories of <ref type="work" corresp="LIT3194Susanna">Susanna</ref>, 
+               <ref type="work" corresp="LIT3195Bel">Bel and the Dragon</ref>, 
+               the <ref type="work" corresp="LIT2250salota">Prayer of Azariah and the Three Holy Children</ref>, which belong to the 
+               so-called deuterocanonical books, or to a text for which this is unknown.</p>
          </abstract>
          <langUsage><language ident="en">English</language></langUsage>
       </profileDesc>
@@ -49,7 +51,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
-         <ab/>
+         <div type="bibliography">
+            <listRelation>
+               <relation name="saws:formsPartOf" active="LIT1259Daniel" passive="LIT3529Daniel"/>
+            </listRelation>     
+         </div>
       </body>
    </text>
 </TEI>
--- a/1001-2000/LIT1259Daniel.xml
+++ b/1001-2000/LIT1259Daniel.xml
@@ -34,6 +34,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
+         <abstract>
+            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the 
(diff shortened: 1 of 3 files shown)


1001-2000/LIT1259Daniel.xml
```diff
@@ -34,6 +34,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
       </encodingDesc>
       <profileDesc>
+         <abstract>
+            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the 
+               Dragon, the Prayer of Azariah and the Three Holy Children, which belong to the so-called deuterocanonical books, or to a or to a 
+               text for which this is unknown.</p>
+         </abstract>
          <langUsage><language ident="en">English</language></langUsage>
       </profileDesc>
       <revisionDesc>
```
eu-genia, line 44: I think we need a `relation `here that it may form part of `3529`

3001-4000/LIT3529Daniel.xml
```diff
@@ -37,6 +33,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
+         <abstract>
+            <p>Biblical Book of Daniel, including the so-called deuterocanonical books with the stories about Susanna, Bel and the dragon, 
+               the Prayer of Azariah and the Three Holy Children.</p>
```
eu-genia, line 38: whole-part relations could be added here or in the corresponding records

eu-genia (commented):
thank you, very helpful

3001-4000/LIT3529Daniel.xml
CarstenHoffmannMarburg: How can I refer to the other ID in relations? I cannot find a relation called "whole-part". I could probably use "saws:contains", but that would require defining the deuterocanonical books as independent IDs. Or do you think, it is sufficient to define LIT1259Daniel to be a part of LIT3529Daniel. This could be misleading, because as far as I know the deuterocanonical books are later additions to the original book of Daniel.

eu-genia: of course the deuterocanonical books must - and already have had - their own IDs, since actually the two Daniel version have always been disambiguated through the title itself 

LIT3529Daniel (with the `title  `"Daniel, Book of (includes Daniel,  Susanna and Bel and the Dragon)" was initially created as the container for

LIT1259Daniel
LIT3194Susanna 
LIT3195Bel
LIT2250salota = LIT1828Mahale#ThreeYouths1 

I checked and they seem to already have the relation in place, but can be re-checked


There are numerous whole-part relations for various degrees of certainty
https://betamasaheft.eu/Guidelines/?id=relations

whole to part 
saws:contains can be used if you are sure that a textual unit **always** contains another textual unit (the latter may circulate also independently). 

part to whole 
`saws:formsPartOf` is used to declare that a defined Textual Unit is `always `part of another one
`ecrm:CLP46i_may_form_part_of` can be used for texts which are `not always` included in another one (and may also circulate independently)
`dcterms:isPartOf` - the more generic relation for parts of larger collections

1001-2000/LIT1259Daniel.xml
```diff
@@ -34,6 +34,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
+         <abstract>
+            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the
```
thea-m, line 38: General

1001-2000/LIT1259Daniel.xml
```diff
@@ -34,6 +34,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
+         <abstract>
+            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the 
+               Dragon, the Prayer of Azariah and the Three Holy Children, which belong to the so-called deuterocanonical books, or to a or to a
```
thea-m, line 39: or to a or to a double :)

thea-m: I think it is also useful to mark up the names of works in abstracts, so that users can find the corresponding page immediately without searching

thea-m (changes requested):
Great, thank you!

1001-2000/LIT1259Daniel.xml
CarstenHoffmannMarburg: Do you think, I should point to LIT3529Daniel in this abstract or to a wikidata (Q80115)?

thea-m: I am not sure if the wikidata reference is resolved (if not, it is not necessarily needed, the references to the other Bm entries are enough, I think)

3001-4000/LIT3529Daniel.xml
```diff
@@ -34,8 +34,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <abstract>
-            <p>Biblical Book of Daniel, including the so-called deuterocanonical books with the stories about Susanna, Bel and the dragon, 
-               the Prayer of Azariah and the Three Holy Children.</p>
+            <p><ref type="work" corresp="wd:Q80115">Biblical Book of Daniel</ref>, including the so-called deuterocanonical books with
```
eu-genia, line 37: above in 1259 you use the same `wikidata ID `for the book of Daniel proper. if the use above is correct, remove `wd `here, as it only multiplies the confusion, the same ID may not be used for the two different constellations of texts

1001-2000/LIT1259Daniel.xml
```diff
@@ -35,9 +35,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <abstract>
-            <p>Generic ID for the biblical Book of Daniel, referring to a recension that does not include the stories of Susanna, Bel and the 
-               Dragon, the Prayer of Azariah and the Three Holy Children, which belong to the so-called deuterocanonical books, or to a or to a 
-               text for which this is unknown.</p>
+            <p>General ID for the biblical <ref type="work" corresp="wd:Q80115">Book of Daniel</ref>, referring to a recension that does
```
eu-genia, line 38: judging by the wikidata structure, ID q80115 stands for `3529`, not for `1259`

**Q218087 and Q16930210 have the part of relation to q80115** 

to avoid unnecessary confusion, unless you are 100% clear about it, remove wd.

now you have inserted in the abstract the same wikidata id for both 1259 and 3529

1001-2000/LIT1259Daniel.xml
```diff
@@ -49,7 +51,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
-         <ab/>
+         <div type="bibliography">
+            <listRelation>
+               <relation name="saws:formsPartOf" active="LIT1259Daniel" passive="LIT3529Daniel"/>
```
eu-genia, line 56: this is correct if it is ALWAYS part of.
if it may have circulation without the other texts then you need the ecrm relation

eu-genia (changes requested):

3001-4000/LIT3529Daniel.xml
thea-m: Yes, I think LIT2271salota should also be added among the relations here.
LIT4300PrayerAzariah has only been used once in BDLaethd3 to encode "Prayer of Azariah and the Song of the Three Holy Children...corresponds with LXX Daniel 3:28-88" (= LIT2250salota+LIT2271salota+LIT1566hababa). I would propose to delete it, as it is confusing matters unnecessarily.

eu-genia (approved):

thea-m (approved):

PR #1053, opened 2024-02-07
Ephr hos 7.02.2024 mk

MarcinKrawczuk:
I have added incipits so that the homilies can be distinguished quickly.


thea-m (approved):
Great, thank you

PR #1056, opened 2024-02-13
Addisie maftəḥe śǝrāy

Addisie:


1001-2000/LIT1824Mafteh.xml
```diff
@@ -77,6 +78,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <relation name="ecrm:CLP46i_may_form_part_of" active="LIT5139MagicPrayer" passive="LIT1824Mafteh"></relation>
 
          </listRelation>
+         <div type="edition">
+            <note>The following text part description is based on the manuscript <ref type="mss" corresp="MEAG002"/>which is incomplete.</note>
+            <div type="textpart" subtype="chapter" xml:id="Monday">
+               <label>Prayer for Monday</label>
```
DenisNosnitsin1970, line 84: 1) If you want to add a structure or text to a work record, you should take a full text witness (here, it  stops on Tuesday) in a way representative for the tradition. 
2) I have some doubts as to if the text is really structured according to the days of the week. I have never seen such a Meftahe seray. (if the means for this are notes in the upper margin, they do not necessarily reflect the structure)
3) Unless it is not done in one or another way, I suggest to add to LIT1824Mafteh the status of "general record". The collections with this title are of unstable composition, not much studied (possibly an issue can be done to discuss this).

DenisNosnitsin1970 (changes requested):

1001-2000/LIT1824Mafteh.xml
```diff
@@ -77,6 +78,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <relation name="ecrm:CLP46i_may_form_part_of" active="LIT5139MagicPrayer" passive="LIT1824Mafteh"></relation>
 
          </listRelation>
+         <div type="edition">
+            <note>The following text part description is based on the manuscript <ref type="mss" corresp="MEAG002"/>which is incomplete.</note>
```
eu-genia, line 82: I would not add edition here at all, use it just in the manuscript for now. Edition for text must be from an edited text or at least from a best witness. 




VERY VERY IMPORTANT: never ever use diacritics or any non-ASCII symbols in any file names, folder names, branch names, or xml:IDs. The name of this branch is a problem.

1001-2000/LIT1824Mafteh.xml
```diff
@@ -78,15 +79,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <relation name="ecrm:CLP46i_may_form_part_of" active="LIT5139MagicPrayer" passive="LIT1824Mafteh"></relation>
```
eu-genia, line 79: `active `must always be the current record 

so relation should be `saws:contains` 

(or you can put the relation as it is now under `LIT5139MagicPrayer`)

eu-genia (changes requested):

CarstenHoffmannMarburg (approved):

1001-2000/LIT1824Mafteh.xml
eu-genia: (and added to LIT5139MagicPrayer)

that is 
either under LIT5139MagicPrayer we have the relation 
`<relation name="ecrm:CLP46i_may_form_part_of" active="LIT5139MagicPrayer" passive="LIT1824Mafteh"></relation>
`

or here in the LIT1824Mafteh record we have
` <relation name="saws:contains" active="LIT1824Mafteh" passive="LIT5139MagicPrayer"></relation>`

the current record is always the `active `member

Addisie:
All changes were carried out: https://github.com/BetaMasaheft/Works/pull/1065/files
                                               https://github.com/BetaMasaheft/Works/pull/1069

PR #1068, opened 2024-02-28
Create LIT6943PredicationsChr

CarstenHoffmannMarburg:
I created a record according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2490.

### DIFF
--- a/new/LIT6943PredicationsChr.xml
+++ b/new/LIT6943PredicationsChr.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">Predictions about the birth of Christ made by the Prophets</title>
+                <title xml:lang="en" xml:id="t1">Predictions about the birth of Christ made by the Prophets</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="CH"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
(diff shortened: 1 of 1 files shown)


new/LIT6943PredicationsChr.xml
```diff
@@ -0,0 +1,67 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-02-28">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129i_is_subject_of" active="LIT6943PredicationsChr" passive="PRS5684JesusCh"/>
```
DenisNosnitsin1970, line 57: I would add a relation to the ms where the text is attested.

DenisNosnitsin1970 (changes requested):

new/LIT6943PredicationsChr.xml
eu-genia: Denis there is already the `witness `in `ListWit`, we use `witness` for attestations of texts

new/LIT6943PredicationsChr.xml
```diff
@@ -0,0 +1,67 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6943PredicationsChr" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Predications about the birth of Christ made by the Prophets</title>
```
eu-genia, line 8: en

new/LIT6943PredicationsChr.xml
```diff
@@ -0,0 +1,67 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="ecrm:P129i_is_subject_of" active="LIT6943PredicationsChr" passive="PRS5684JesusCh"/>
+                </listRelation>
+            </div><!---->
+            <div type="edition" subtype="textsubunit">
```
eu-genia, line 60: it should be 
```
<div type="edition" xml:lang="gez">
 <div type="textpart">
 <ab>your text here</ab>
</div>
</div>
```
there is potentially the edition of the entire text, and nested there is a textpart. textsubunit would be a clearly marked chapter or paragraph, I do not think this is the case.

new/LIT6943PredicationsChr.xml
```diff
@@ -0,0 +1,67 @@
...
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianContent"/>
+                    <term key="ChristianLiterature"/>
+                    <term key="OldTestament"/>
```
eu-genia, line 41: OldTestament is for biblical books for the OT, remove

eu-genia (changes requested):

new/LIT6943PredicationsChr.xml
DenisNosnitsin1970: "Predications" is smth different, as far as I can see. I assume Strelcyns title can be taken over.

DenisNosnitsin1970 (changes requested):

PR #1084, opened 2024-04-16
Create LIT6957HassabaHegg.xml

CarstenHoffmannMarburg:
I created a record for a calendar in BLorient13269 as discussed in https://github.com/BetaMasaheft/Manuscripts/pull/2339 and https://github.com/BetaMasaheft/Documentation/issues/2510.

### DIFF
--- a/new/LIT6957HassabaHegg.xml
+++ b/new/LIT6957HassabaHegg.xml
@@ -57,6 +57,26 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <relation name="saws:isDifferentTo" active="LIT6957HassabaHegg" passive="LIT6563HassabHeg"/>
                 </listRelation>
             </div><!---->
+            <div xml:lang="gez" type="edition">
+                <note>The following incipit is taken from <ref type="mss" corresp="BLorient13269"/> as quoted in
+                    <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">51</citedRange></bibl>.</note>
+                <div type="textpart" subtype="incipit">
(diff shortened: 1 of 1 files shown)


new/LIT6957HassabaHegg.xml
```diff
@@ -0,0 +1,82 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT6957HassabaHegg" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Ḥassāba ḥǝgg</title>
```
DenisNosnitsin1970, line 8: It would be better to shape the title section according to the scheme in the guidelines, `<title xml:lang="gez" xml:id="t1">መጽሐፈ፡ ቅዳሴ፡</title>`
         `<title xml:lang="gez" corresp="#t1" type="normalized">Maṣḥafa qǝddāse</title>`
         `<title xml:lang="en" corresp="#t1">Book of the Consecration</title>` 
that is, with Ethiopic and translation, even though approximate.

DenisNosnitsin1970 (changes requested):

new/LIT6957HassabaHegg.xml
CarstenHoffmannMarburg: What do you suggest as translation for Ḥassāba ḥǝgg? Is "Calculation of feasts" an acceptable approximation?

DenisNosnitsin1970: I would translate literally, "Calculation of the law", whateve it might mean; just in case it is completely misleading, I create an additional supplied English title.

eu-genia (approved):

PR #1086, opened 2024-04-18
Etana mogar LIT4013EtanaM

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/2504 We now have EtanaMogar as keyword, added to this record, but it probably has to be split into (three?) separate ones with changes in mss, or renamed as a general record (? https://github.com/BetaMasaheft/Documentation/issues/1289)

### DIFF
--- a/4001-5000/LIT4013EtanaM.xml
+++ b/4001-5000/LIT4013EtanaM.xml
@@ -18,9 +18,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <pubPlace>Hamburg</pubPlace>
             <availability>
                <licence target="http://creativecommons.org/licenses/by-sa/4.0/">
-                  <p>
-                                    This file is licensed under the Creative Commons
-                                    Attribution-ShareAlike 4.0. </p>
+                 This file is licensed under the Creative Commons
+                                    Attribution-ShareAlike 4.0. 
                </licence>
             </availability>
          </publicationStmt>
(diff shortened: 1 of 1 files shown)


eu-genia:
NB there is also NAR0183EtanaMogar used in BLorient570

DenisNosnitsin1970:
It can be used as an ID for this specific qene, but the title should be reshaped. Key word, yes, to indicate one of the types known.

DenisNosnitsin1970:
Well, here all poems can get separaite IDs, since incipits and types are known.

eu-genia:
This is what I meant, the existing ID should point to one ms only, two new IDs must be created

eu-genia:
I have created new records for individual poems formerly collected under LIT4013EtanaM and under NAR0183EtanaMogar. 
LIT5129Qene is now not a general record but used for a specific poem.

Should we delete LIT4013EtanaM and NAR0183EtanaMogar or keep one of these for general record purposes?

eu-genia:
LIT4013EtanaM is still used in BNFet119 
```
<msItem xml:id="ms_i5">
                     <title type="complete" ref="LIT4013EtanaM" xml:lang="gez">ዕጣነ፡ ሞገር፡</title>
                     <textLang mainLang="gez"/>
                  </msItem>
```
I guess we do not have the images to provide a disambiguation/longer title, so some generic ID is needed, but we could also use the NAR one (not used anywhere after this PR https://github.com/BetaMasaheft/Manuscripts/pull/2407)

PR #1088, opened 2024-04-19
CreateLIT6967PrayerPassion

CarstenHoffmannMarburg:
I created a record for a "Prayer on the Passion of Jesus Christ, which Jesus revealed to Matilda, Bridget and Elizabeth" according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2509. Unfortunately, I was not able to check the manuscript text in KopCodEtiopAdd13, because the record was not available, but there was an error note: "HTTP ERROR 500 ..."

### DIFF
--- a/new/LIT6967PrayerPassion.xml
+++ b/new/LIT6967PrayerPassion.xml
@@ -1,0 +1,67 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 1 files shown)


eu-genia:
The Copenhagen descriptions do not follow the schema and most of them will not be displayed unless the encoding is corrected.
So if you want to see them you either check the XML or fix it (done now for Ms 13, where the error was clearly stated in the 
`HTTP ERROR 500 javax.servlet.ServletException: javax.servlet.ServletException: An error occurred while processing request to /exist/apps/BetMasWeb/manuscripts/KopCodEtiopAdd13/main: exerr:ERROR XPTY0004: The actual cardinality for parameter 1 does not match the cardinality declared in the function's signature: string($arg as item()?) xs:string. Expected cardinality: zero or one, got 3. [at line 2168, column 13, source: /db/apps/BetMasWeb/modules/viewItem.xqm] In function: viewItem:layoutdimensionunit(item()*)`
(multiple dimensions inside of one layout element)

If you see an Error like this you always can either fix the XML yourself or ask someone to fix it. We have many such cases with records that were transformed from another source.

eu-genia:
We do not have images for Copenhagen but Denis probably does

eu-genia (approved):

eu-genia (approved):

PR #1101, opened 2024-05-07
Update LIT1544Gebrah (add chapters as Sub-IDs)

CarstenHoffmannMarburg:
I added some chapters to the record of LIT1544Gebrah as discussed in https://github.com/BetaMasaheft/Documentation/issues/2531. I also added some titles to chapters already existing. However, I saw, that there is no subdivision for each hour of the day or of the night (except for Friday), as it is regularly in manuscripts. If you agree, I would like to add also chapters for the other days of the week, unless the place of some readings may vary in some manuscripts.

### DIFF
--- a/1001-2000/LIT1544Gebrah.xml
+++ b/1001-2000/LIT1544Gebrah.xml
@@ -47,12 +47,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </textClass>
       </profileDesc>
       <revisionDesc>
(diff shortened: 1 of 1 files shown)


eu-genia (approved):
thank you!

CarstenHoffmannMarburg:
I identified each chapter, that I recorded in a div in ESqmb001, which was the only witness of LIT1544Gebrah, that was available to me.

thea-m:
Thank you, I think it makes sense to include the subdivisions for the additional chapters in this PR as well

thea-m (approved):

PR #1103, opened 2024-05-13
Add Edition to Malkəʾa ʾArsimā

smaugustine:
This is to add the TEI encoded version of the edition presented in the new volume of Aethiopica.

### DIFF
--- a/2001-3000/LIT2900RepCh181.xml
+++ b/2001-3000/LIT2900RepCh181.xml
@@ -70,6 +70,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <ptr target="bm:Dickinson2023Arsima"/>
           </bibl>
         </listBibl>
+
+        <listBibl type="translation">
+          <bibl>
+            <ptr target="bm:Dickinson2023Arsima"/>
+          </bibl>
+        </listBibl>
       </div>
     
       <div type="edition" xml:lang="gez">
@@ -217,7 +223,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
         <app>
           <lem wit="#A">
 
-            <ab><choice><corr resp="AD">ተረፈ፡ መልክእ፡</corr><sic/></choice></ab>
+            <ab><supplied resp="AD" reason="explanation">ተረፈ፡ መልክእ፡</supplied></ab>
 
             <lg type="stanza" n="A">
               <l n="1"><pb edRef="#A" n="169v"/><cb edRef="#A" n="a"/>ሰላም፡ ለኪ፡ አርሲማ፡ ኅሪት፨</l>
@@ -419,7 +425,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
           <l n="5">In the place where the offering of Abel rose before God.</l>
         </lg>
 
-        <ab>Tarafa Malkəʾ</ab>
+        <ab>[Tarafa Malkəʾ]</ab>
         
         <lg type="stanza" n="A">
           <l n="1">Salutation to you, chosen Hripsime:</l>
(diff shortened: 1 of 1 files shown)


smaugustine:
Should I add the translation as well? Can it be encoded similarly?

I also would like to encode the rhyme. `<rhyme>` seems not to be allowed (at least within `<l>`) according to the schema. Does anyone have a preference for how the rhyme of lines/stanzas should be encoded?

eu-genia:
thank you, I will have to first adjust the schema, for the moment we do not even have lg in the definitions or in the xqm
https://github.com/BetaMasaheft/Schema/blob/master/tei-betamesaheft.xml

if we need `lg `and `type `and `rhyme` (and guidelines to describe this) please open an issue with requirements to implement, thank you!

so far we have used `div type="stanza"` (and `lb`)

see e,g, https://betamasaheft.eu/works/LIT3055RepCh335/text

we can probably change it but it must be discussed, introduced into the guidelines, and old records must be then revised accordingly

smaugustine:
`lg` was meant to have been implemented (I'm not sure which level and in precisely which way) from [this issue](https://github.com/BetaMasaheft/Documentation/issues/1483) and is included in the BM guidelines ([example 8 here](https://betamasaheft.eu/Guidelines/?id=edition)).

Rhyme can always be left for another day, especially with such straightforward cases as these, where it could be inserted simply with RegEx, even. I think, at least for this class of poetry, rhyme will become more significant when mixed-rhyme stanzas are encoded, but this is my idea for a project years in the future and need not be a pressing concern for now if it is not (yet) needed by others.

smaugustine:
The use of `lg` and `l` for poetry as opposed to `div` and `lb` [has been the case for some time](https://github.com/search?q=repo%3ABetaMasaheft%2FWorks%20lg&type=code), unless there is something I've misunderstood.

eu-genia:
it has never been implemented in the xquery, or rather, at that time the views were produced through xslt and anything worked, just before leaving, however, Pietro launched a new release where all views are generated by xquery, and `lg `do not work there, you can easily see that for none of the records where `lg `has been used the text display works in the app

eu-genia:
the schema was never updated either, which is the main reason the xqm ignores the element and throws an error

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/2537

eu-genia:
ok the xquery is fixed, schema is OK as long as we do not add rhyme and other extra things, please shout if you discover any bugs (I had never checked `lg `containing records previously so did not see the error in the xquery)

2001-3000/LIT2900RepCh181.xml
```diff
@@ -1,73 +1,286 @@
...
+            <citedRange unit="item">181</citedRange>
+          </bibl>
+
+          <bibl>
+            <ptr target="bm:Dickinson2023Arsima"/>
+          </bibl>
+        </listBibl>
+      </div>
+    
+      <div type="edition">
```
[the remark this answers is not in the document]
eu-genia: (always language for editions - or translations which of course also may be added in an extra div)

smaugustine:
Is it strictly necessary to add `corresp` for the translation? At which level would be acceptable?

eu-genia:
It is not necessary, especially since at the moment we have not implemented anything for parallel or linked display. 
We might, in the future, but then we can see if the need arises.
I guess it could be done at any level, one would have to provide xml:id to the stanzas to link them, but it is OK as it is for now, thank you!

eu-genia (approved):

2001-3000/LIT2900RepCh181.xml
```diff
@@ -1,73 +1,484 @@
...
+          <bibl>
+            <ptr target="bm:Chaine1913Rep"/>
+            <citedRange unit="item">181</citedRange>
+          </bibl>
+
+          <bibl>
+            <ptr target="bm:Dickinson2023Arsima"/>
+          </bibl>
+        </listBibl>
+      </div>
```
eu-genia, line 73: I just thought that one could also add the relation 
```
            <listRelation>
                <relation name="ecrm:P129_is_about" active="LIT2900RepCh181" passive="PRS8126Rhipsime"/>
             </listRelation>
```

eu-genia: and one could also add
```
<listBibl type="translation">
          <bibl>
            <ptr target="bm:Dickinson2023Arsima"/>
          </bibl>
        </listBibl>
```

smaugustine: Both came to my mind at one point or another then I promptly forgot about them — let me add them.

PR #1129, opened 2024-07-01
Re merge 1.07.24 mk

MarcinKrawczuk:
So re-merging has worked! I have now tried to do the same but to four records in one branch. I have kept the records mostly as Eugenia left them, just added Ethiopic title in two cases


CarstenHoffmannMarburg (changes requested):

MarcinKrawczuk:
Corrected

wt., 2 lip 2024 o 19:27 Carsten Hoffmann ***@***.***>
napisał(a):

CarstenHoffmannMarburg (approved):

PR #1142, opened 2024-07-11
Update LIT4860Litanies.xml

LeonardBahr:
Hello everyone, I have added a new prayer that is being created (https://github.com/BetaMasaheft/Documentation/issues/2561) to the existing entry "Litanies for the days of the week" (https://betamasaheft.eu/works/LIT4860Litanies/main).


CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg:
I think it would be more convenient to push LIT7016Liton and all the other related changes in a single pull request. It is easier for reviewers to understand what happened. As far as, there are not more than 10 records, it is possible to have several records in one pull request.

LeonardBahr:
Sounds possible. What if we procede with LIT7016 first, and then do the rest, as you suggested, all in one pull request? I would like to finish the working unit of this one first, because it will take some time until I will be able to edit all the others simultaneously as I am busy the next weeks.

eu-genia (approved):

PR #1145, opened 2024-07-13
Lit1804 lifeof corr mk 11.07.24

MarcinKrawczuk:
Added abstract, relation and witness (EMML 1840), expanded bibliography, corrected attribution.

### DIFF
--- a/1001-2000/LIT1804Lifeof.xml
+++ b/1001-2000/LIT1804Lifeof.xml
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
-schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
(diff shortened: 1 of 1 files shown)


1001-2000/LIT1804Lifeof.xml
```diff
@@ -52,6 +52,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
+         <abstract>
+<p>According to the colophon in <ref type="mss" corresp="EMML1840">EMML 1840, fol. 223r</ref> the text was commissioned by the <roleName type="title">ʿaqqābe saʿat</roleName> <ref type="pers" corresp="PRS8538saraqaB">Śaraqa Bǝrhān</ref>.</p>
```
MassimoVilla, line 56: I am not sure fol. 223r can be nested here inside `ref type="mss"`, Is this marking up used elsewhere or allowed in the guidelines?

MassimoVilla (changes requested):

MarcinKrawczuk:
I will not be working on this today until late afternoon/evening. Perhaps
we could just leave here EMML1840 without the fol. and I will transcribe
the colophon and add it to the manuscript description?

pon., 15 lip 2024 o 09:38 MassimoVilla ***@***.***>
napisał(a):

MassimoVilla:
I have changed the reference to fol. 223r adding <locus target="#223r"/>. I think this works

1001-2000/LIT1804Lifeof.xml
CarstenHoffmannMarburg: I think, I have pointed to specific folia in manuscripts like this in other records, as well. I think, it will work.

CarstenHoffmannMarburg (approved):

MassimoVilla (approved):

PR #1170, opened 2024-08-28
Create LIT7048Qene (no. 2)

CarstenHoffmannMarburg:
I struggled with the translation of the title and did not know, which Qene-type it is.

### DIFF
--- a/new/LIT7048Qene.xml
+++ b/new/LIT7048Qene.xml
@@ -6,8 +6,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">ከልአኪ፡ አሮን፡ በበትረ፡ እዴሁ፡ ጥልልት።</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Kalʾa-ki ʾAron ba-batra ʾǝde-hu ṭǝllǝlt (Qǝne)</title>
-                <title xml:lang="en" corresp="#t1">Aron's hand prevented you to be fertile in the offspring (?) ...</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Kalʾa-ki ʾAron ba-batra ʾǝde-hu ṭǝllǝlt (Qǝne, possibly of ʾƎtāna mogar-type)</title>
+                <title xml:lang="en" corresp="#t1">Aron prevented you with the verdant stick of his hand ...</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="CH"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -34,12 +34,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>Qǝne; Type and author are not indicated in the manuscript.</p>
+                <p>Qǝne, possibly of the ʾƎṭāna mogar-type; type and author are not indicated in the manuscript. The poem has 11 verses, but the 
+                    first two verses, which come from another qǝne poem, have a different rhyme.</p>
             </abstract>
             <textClass>
                 <keywords>
                     <term key="Poetry"/>
                     <term key="Qene"/>
+                    <term key="EtanaMogar"/>
                 </keywords>
             </textClass>
             <langUsage>
--- a/new/LIT7048Qene.xml
+++ b/new/LIT7048Qene.xml
@@ -35,7 +35,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <creation/>
             <abstract>
                 <p>Qǝne, possibly of the ʾƎṭāna mogar-type; type and author are not indicated in the manuscript. The poem has 11 verses, but the 
-                    first two verses, which come from another qǝne poem, have a different rhyme.</p>
+                    first two verses, which come from another qǝne poem, have a different rhyme. Division into verses is indicated with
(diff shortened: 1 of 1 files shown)


new/LIT7048Qene.xml
```diff
@@ -0,0 +1,75 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7048Qene" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ከልአኪ፡ አሮን፡ በበትረ፡ እዴሁ፡ ጥልልት።</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Kalʾa-ki ʾAron ba-batra ʾǝde-hu ṭǝllǝlt (Qǝne)</title>
+                <title xml:lang="en" corresp="#t1">Aron's hand prevented you to be fertile in the offspring (?) ...</title>
```
DenisNosnitsin1970, line 10: Aron prevented you with the verdant stick of his hand ... 
"Stick" can be femenine (ወሠረጸት፡ በትር...), the allusion he is to the stick of Aron that shoot sprouts (Num 17).

new/LIT7048Qene.xml
```diff
@@ -0,0 +1,75 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7048Qene" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ከልአኪ፡ አሮን፡ በበትረ፡ እዴሁ፡ ጥልልት።</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Kalʾa-ki ʾAron ba-batra ʾǝde-hu ṭǝllǝlt (Qǝne)</title>
```
DenisNosnitsin1970, line 9: We can indicate: Qǝne, possibly of etana mogar type. 
Or: of undefined type. 
In the abstract we can comment that the type is possibly etana mogar.
This has 11 verses, but a different rhyme of the first two I cannot explain. Perhaps 
 it follows a special type or (regional) subtype, but I do not see it for the moment. By the way, as a reference I use Guidi's "Qene, or inni abissini", 1901, with a list of types  and examples (I will circulate a scan). There are some other survey of qene types, Guidi is good and precise. We can gradually collect a bibliography pn qene, this will not be too much (but there is an unknown number of Ethiopian editions).

new/LIT7048Qene.xml
```diff
@@ -0,0 +1,75 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7048Qene" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ከልአኪ፡ አሮን፡ በበትረ፡ እዴሁ፡ ጥልልት።</title>
```
DenisNosnitsin1970, line 8: Indeed, to Jonas idea, I see that ከልአኪ፡ አሮን፡ በበትረ፡ እዴሁ፡ ጥልልት። ወሕዝቅኤል፡ ጽባሐዊ፡ ዓቀመኪ፡ በኆኅት። was used in another qene.

DenisNosnitsin1970 (changes requested):

new/LIT7048Qene.xml
CarstenHoffmannMarburg: Do we keep this text as one qene poem nevertheless, or should we split into two? Should I refer to the other qene in a relation and in the abstract?

DenisNosnitsin1970 (changes requested):

new/LIT7048Qene.xml
```diff
@@ -0,0 +1,77 @@
...
+                <title xml:lang="gez" corresp="#t1" type="normalized">Kalʾa-ki ʾAron ba-batra ʾǝde-hu ṭǝllǝlt (Qǝne, possibly of ʾƎtāna mogar-type)</title>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: You are right, we do not seperate possessive suffixes in translitterations (as in "Salām laka").

new/LIT7048Qene.xml
DenisNosnitsin1970: We keep it as one. A reference to other qene in abstract is ok, we will see if we can collect more cases like this (I expect we will have some), to see if it can be encoded. It is indeed curious if the first two verses, having been borrowed, keep the rhyme of the source. What would qene masters say about this? (the rule is that qenes should be fully original, but the reality is a different thing, there are treatise offering ready verses for qenes) :)

DenisNosnitsin1970 (approved):

PR #1185, opened 2024-09-06
2024 09 06 jk create marian prayer uub o etiop 10

karljonaskarlsson:
Created LIT7065 and added relations to two other prayers with similar beginnings to differentiate them, all according to https://github.com/BetaMasaheft/Documentation/issues/2625#issuecomment-2333797110


CarstenHoffmannMarburg:
Thank you very much! Well done! I think the three can easily be told apart by the incipits you gave. It would be nice if you could name some witnesses in the source citation.

CarstenHoffmannMarburg (approved):

nafisa-valieva (approved):

PR #1186, opened 2024-09-06
Create LIT7066NiceneCreedAmh.xml

karljonaskarlsson:
Created file based on https://github.com/BetaMasaheft/Documentation/issues/2624#issuecomment-2328937856

### DIFF
--- a/new/LIT7066NiceneCreedAmh.xml
+++ b/new/LIT7066NiceneCreedAmh.xml
@@ -1,0 +1,68 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7066NiceneCreedAmh" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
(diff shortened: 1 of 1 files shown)


new/LIT7066NiceneCreedAmh.xml
```diff
@@ -0,0 +1,69 @@
...
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <p/>
```
CarstenHoffmannMarburg, line 24: Add witness UppEt10 to source desc.

new/LIT7066NiceneCreedAmh.xml
nafisa-valieva, line 1: (Amharic) in title looks a bit strange, if I may have my opinion. We do not write (Geez)  :-)

CarstenHoffmannMarburg: It is not consistent, but for practical reasons very useful!

new/LIT7066NiceneCreedAmh.xml
karljonaskarlsson: I thought we used <listWit> for listing the manuscripts used in an edition?

karljonaskarlsson: Ah, I understand that the witness used for our edition (= incipit, in this case) can be listed as a witness, so I will add UppEt10, as you suggested. Thank you!

eu-genia (approved):

CarstenHoffmannMarburg (approved):

PR #1199, opened 2024-09-19
Smaller corrections of LIT6707 and LIT6708

CarstenHoffmannMarburg:
I have seen, that there is a minor problem with the representation on the website having Incipitincipit or Explicitexplicit. I think, this happened, because <label>Incipit</label> was put into the record. I deleted <label> and I hope, that this problem is solved with this change.

Screenshot 2024_Incipitincipit


eu-genia:
Thank you, of course your changes would remove it, but I just wonder if this case actually means that it is not the XML that must be adjusted but the visualization in the app.

Currently so it seams we get printed for each `textpart`

`label+subtype` in small italics
`label `in big roman

removing the label in the xml will of course remove it in first case but also in the second case

If we never need label when we have a subtype this is OK for me but I guess there are cases where we might need both, and then it is to be decided what and how must be printed.

eu-genia (approved):

PR #1201, opened 2024-09-20
2024 09 19 update salota etan

karljonaskarlsson:
Updates based on comments in https://github.com/BetaMasaheft/Manuscripts/pull/2619

Added incipits in LIT2252salota. Added a section in LIT1960Mashaf with references to the different text parts of LIT2252salota (as attested in UppEt20, see https://github.com/BetaMasaheft/Manuscripts/pull/2619)

### DIFF
--- a/1001-2000/LIT1960Mashaf.xml
+++ b/1001-2000/LIT1960Mashaf.xml
@@ -53,6 +53,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
          <change who="DR" when="2017-03-27">Minor correction</change>
          <change who="DR" when="2017-04-04">Added keywords</change>
          <change when="2018-03-27" who="ES">added prayers/textparts from the Maṣḥāfa nǝfq in ESagm009</change>
+         <change when="2024-09-20" who="JK">added prayers/textparts from UppEt20</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
--- a/1001-2000/LIT1960Mashaf.xml
+++ b/1001-2000/LIT1960Mashaf.xml
@@ -122,7 +122,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                
                <div type="textpart" xml:id="PrayerIncense" corresp="LIT2252salota">
                   <label>Prayers of the Incense (collection of prayers)</label>
-                  <!--This section follows the attestation in UppEt20. It is different from <ref corresp="#Incense"/>-->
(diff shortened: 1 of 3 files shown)


1001-2000/LIT1960Mashaf.xml
```diff
@@ -114,11 +115,55 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                   </div>
+               </div>
                
                
-               </div>
                
                
+               <div type="textpart" xml:id="PrayerIncense" corresp="LIT2252salota">
+                  <label>Prayers of the Incense (collection of prayers)</label>
+                  <!--This section follows the attestation in UppEt20. It is different from <ref corresp="#Incense"/>-->
```
CarstenHoffmannMarburg, line 125: This information should be given in a note, which is visible to the user.

karljonaskarlsson: In other parts of the record, it seems that this kind of information is given in "ausgeklammerten" comments

karljonaskarlsson: (To be honest, the entire LIT1960Mashaf is quite messy. I can follow your suggestion for the part that I added: it will be unsystematic, but maybe it's rather the other parts that should be changed 👍 )

eu-genia: The commenting out depends on whether this comment is just for you as the encoder, as a temporary reminder to encode something later, because you do not know whether and how to encode - then yes comment out, it is then not visible for anyone, not searchable, and will not be also present in the expanded TEI XML on the server.

if you want to put it as if it were a footnote, or a note in brackets, then it must be properly encoded as a `note `or similar element.

note that references to other elements in the same record should have the target `<ref target="#Incense">Prayers of the Incense</ref>`

This information can also be provided as a relation in LIT2252salota (`<relation name="saws:isDifferentTo" active="LIT2252salota" passive="LIT1960Mashaf#Incense">`)

karljonaskarlsson: Thanks! My intention was to note this down for the future person who will clean up LIT1960Mashaf (based on a better understanding on this work than mine), but I turned it into a general note instead. The information might be relevant in general, as noted by Carsten.

karljonaskarlsson: I added the relation in LIT2252salota according to your suggestion. Thank you! :)

nafisa-valieva (approved):

CarstenHoffmannMarburg (approved):

eu-genia (approved):

PR #1204, opened 2024-09-22
Lit2365 storyo corr mk 20.09.24

MarcinKrawczuk:
I would to prefer to leave the English title, the incipit of the text does not provide anything "title-like" enough in my opinion. Also, no folio numbers in the EMDA ms. deliberately, it is not paginated.


CarstenHoffmannMarburg:
To my opinion, ነገር፡ ዘበእንተ፡ መልከ፡ ጼዴቅ፡ or ነገር፡ ዘበእንተ፡ መልከ፡ ጼዴቅ፡ ዘሞአ፡ ለሞት፡, as given in the incipit, would serve well as title.

CarstenHoffmannMarburg (approved):

PR #1205, opened 2024-09-23
Update LIT4327Basamay.xml

karljonaskarlsson:
Added reference to LIT1960 based on https://github.com/BetaMasaheft/Manuscripts/pull/2619#discussion_r1768443178

### DIFF
--- a/4001-5000/LIT4327Basamay.xml
+++ b/4001-5000/LIT4327Basamay.xml
@@ -45,6 +45,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
+         <div type="bibliography">
+            <listRelation>
+               <relation name="ecrm:CLP46i_may_form_part_of" active="#LIT4327Basamay" passive="LIT1960Mashaf"/>
+            </listRelation>
+         </div>
          <div type="edition">
             <div type="textpart" subtype="incipit" xml:lang="gez">
                <ab>በሰማይ፡ የሀሉ፡ ልብክሙ፡ እወ፡ የሀሉ፡ በሰማይ፡ ልብነ፡</ab>


LeonardBahr:
This (= CAe 4327?) looks like an excerpt from the beginning of Qǝddāse ʾƎgziʾ = the Anaphora of Our Lord Jesus Christ, found, e.g., in MKL-008, 17th c., on fols. 44vb–49vb.

LeonardBahr:
Or in this manuscript from the 20th c., on fols 94r–99r: : https://betamasaheft.eu/EMIP01314.

4001-5000/LIT4327Basamay.xml
```diff
@@ -45,6 +45,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
       <body>
+         <div type="bibliography">
+            <listRelation>
+               <relation name="ecrm:CLP46i_may_form_part_of" active="#LIT4327Basamay" passive="LIT1960Mashaf"/>
```
eu-genia, line 50: remove `#`

eu-genia: @LeonardBahr is of course right, we should add also the part of relation to LIT1100Anapho 


this is always a tricky decision whether we describe the whole chain of relations

actually this is part of LIT1100Anapho (Anaphora of our Lord) which is part of LIT3189Anapho (14 Anaphoras) which is part of LIT1960Mashaf (Ser'ata qeddase)

karljonaskarlsson: The manuscript I was encoding (UppEt20) also contains the Anaphora of Our Lord in a different place.

However, I think there is an explanation for what's going on in the manuscript. This portion (= CAe 4327) is namely not found in the "real" attestation of the Anaphora of Our Lord, where one, instead, finds a secondary note (in an originally blank space) saying: ይ፡ ዲ፡ በሰማይ፡ የሃሉ፡ . So presumably, this part was left out at first (for some reason), and then the scribe felt a need to add it, but then didn't have enough space next to the actual Anaphora of Our Lord. And so, they added it at the end of the manuscript instead (which is where we find "CAe 4327" now).

In terms of encoding, I'm not sure how this is best handled. Maybe "CAe 4327" in UppEt20 is better defined as a part of the Anaphora of Our Lord?

UppEt20, in that case, is no witness to the independent circulation of "CAe 4327", but maybe there are other witnesses…

eu-genia:
At least in two of the witnesses registered on Bm it seems to be independent.
I notice also that in BDLaethf1 we have both LIT1100Anapho and LIT4327Basamay, similar to this case

karljonaskarlsson:
@eu-genia , can this be merged?

eu-genia (approved):

PR #1207, opened 2024-09-25
2024 09 25 jk qene mamhera

karljonaskarlsson:
Created record for a qene attested in Kolmodin's collection (as part of creating a description of this manuscript): https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0446&pid=alvin-record:280997

### DIFF
--- a/new/LIT7072QeneMamhera.xml
+++ b/new/LIT7072QeneMamhera.xml
@@ -6,7 +6,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Mamǝhǝra zǝntu ʿālam, ʾAwǝsṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro… (Qǝne of the XXXXXXXXX-type)</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Mamǝhǝra zǝntu ʿālam, ʾAwǝsṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro… (Qǝne of unknown type)</title>
                 <title xml:lang="en" corresp="#t1">The teacher of this world, ʾEwosṭātewos, came to the mountain of Armenia to praise Your Son…</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="JK"/>
@@ -34,8 +34,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>Qǝne without type indication in the manuscript. In the first line, it refers to <persName ref="PRS3934Ewostate"/>.
-                In the manuscript, it is attributed to <persName ref=""></persName>.</p>
+                <p><foreign xml:lang="gez">Qǝne</foreign> without type indication in the manuscript. In the first line, it 
+                    refers to <persName ref="PRS3934Ewostate"/>.
+                    In the manuscript, it is attributed to <persName ref="PRS14574GabraLeul" cert="high" corresp="SD"/>.</p>
             </abstract>
             <textClass>
                 <keywords>
--- a/new/LIT7072QeneMamhera.xml
+++ b/new/LIT7072QeneMamhera.xml
@@ -5,9 +5,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Mamǝhǝra zǝntu ʿālam, ʾAwǝsṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro… (Qǝne of unknown type)</title>
-                <title xml:lang="en" corresp="#t1">The teacher of this world, ʾEwosṭātewos, came to the mountain of Armenia to praise Your Son…</title>
+                <title xml:lang="gez" xml:id="t1">መምህረ፡ ዝንቱ፡ ዓለም፡ <choice><sic>አ</sic><corr resp="JK">ኤ</corr></choice>ዎስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Mamhǝra zǝntu ʿālam, ʾEwosṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro (Qǝne of unknown type)</title>
+                <title xml:lang="en" corresp="#t1">The teacher of this world, ʾEwosṭātewos, came to the mountain/monastery of Armenia to praise Your Son…</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="JK"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -34,8 +34,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p><foreign xml:lang="gez">Qǝne</foreign> without type indication in the manuscript. In the first line, it 
-                    refers to <persName ref="PRS3934Ewostate"/>.
+                <p><foreign xml:lang="gez">Qǝne</foreign> without type indication in the manuscript. It has eight lines, meaning
+                    that it can be classified as a <foreign xml:lang="gez">mawaddəs qəne</foreign> according to
+                    <bibl><ptr target="bm:Guidi1900Qene"/></bibl>. 
+                    In the first line, it refers to <persName ref="PRS3934Ewostate"/>.
                     It is attributed to <persName ref="PRS14574GabraLeul" cert="high" corresp="SD"/> in the manuscript.</p>
             </abstract>
             <textClass>
@@ -43,6 +45,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <term key="ChristianLiterature"/>
                     <term key="Poetry"/>
                     <term key="Qene"/>
+                    <term key="Mawaddes" cert="medium"/>
                 </keywords>
             </textClass>
             <langUsage>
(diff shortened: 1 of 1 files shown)


new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7072QeneMamhera" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Mamǝhǝra zǝntu ʿālam, ʾAwǝsṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro… (Qǝne of unknown type)</title>
```
eu-genia, line 9: not Mamhǝra ?

eu-genia (approved):

new/LIT7072QeneMamhera.xml
karljonaskarlsson: It's a question of contention, I think! Etymologically, one would expect mamǝhǝr (cf. magaśśǝṣ), but in the current pronunciation it's without schwa

CarstenHoffmannMarburg: I think, we should be consistent to other Bm records. Some work records (13), many PRS records (35) and very many manuscript records (217) contain the transcription mamhər. For this reason, we should use this transcription, even though it might be etymologically incorrect.

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+                <title xml:lang="en" corresp="#t1">The teacher of this world, ʾEwosṭātewos, came to the mountain of Armenia to praise Your Son…</title>
```
CarstenHoffmannMarburg, line 10: I wonder, if this can also be translated as "came to **a** mountain **in** Armenia" or "came to **a monastery in** Armenia".

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="JK" when="2024-09-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT7072QeneMamhera" passive="UppQ1524o"/>
```
eu-genia, line 61: maybe the relation isAbout can be helpful, then the hymn will be visualized under Ewostatewos among the works dedicated to him

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+                </listRelation>
+            </div>
+            <div type="edition" xml:lang="gez">
+                <note>This edition is based on <ref type="mss" corresp="UppQ1524o"/> on fols 40r, 41r. Division into verses is indicated with ፤ .
+                    The end of the poem is indicated with ። .</note>
+                <ab>
+                    <l n="1">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</l>
```
CarstenHoffmannMarburg, line 68: It is written አዎስጣቴዎስ፡ here.

new/LIT7072QeneMamhera.xml
eu-genia: ecrm:P129_is_about

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+                    <l n="2">አምጣነ፡ ቦቱ፡ ተስእለ፡ ዘያስተፌስህ፡ አእመሮ፤</l>
+                    <l n="3">ወከመ፡ ዘንቃህ፡ እምንዋም፡ ተንስአ፡ ብኃይሉ፡ ወቀተለ፡ ጸሮ፤</l>
+                    <l n="4">ማዕረረ፡ ሞት፡ ኢያሪኮ፡ ቦቲኮ፡ ዓጽፍከ፡ 
+                        እግዚ<choice><orig>እ</orig><corr>አ</corr></choice>፡
```
CarstenHoffmannMarburg, line 72: In this case, the እ was crossed out by the scribe and a ኣ (or አ) was written above. I think, this should be coded as:
 `<del rend="overstrike">እ</del><add place="above">ኣ</add>`.

The same with ብ / ቢ i.e.: `<del rend="overstrike">ብ</del><add place="above">ቢ</add>`

Otherwise it will look like you have corrected it.

new/LIT7072QeneMamhera.xml
CarstenHoffmannMarburg: It is possibly the monastery called ደብር፡ ለስርአተ፡ ሌዊ፡, which is mentioned in the text in line 76.

CarstenHoffmannMarburg (changes requested):

new/LIT7072QeneMamhera.xml
CarstenHoffmannMarburg: If it has eight verses, I would define it as mawaddǝs type according to Guidi 1901.

karljonaskarlsson: > I think, we should be consistent to other Bm records. Some work records (13), many PRS records (35) and very many manuscript records (217) contain the transcription mamhər. For this reason, we should use this transcription, even though it might be etymologically incorrect.

Can we add this to the list of dubious spellings? https://betamasaheft.eu/Guidelines/?id=dubious-spelling @eu-genia 

Or would it be preferable to have a bigger discussion about it before?

eu-genia: > > I think, we should be consistent to other Bm records. Some work records (13), many PRS records (35) and very many manuscript records (217) contain the transcription mamhər. For this reason, we should use this transcription, even though it might be etymologically incorrect.
> 
> Can we add this to the list of dubious spellings? https://betamasaheft.eu/Guidelines/?id=dubious-spelling @eu-genia
> 
> Or would it be preferable to have a bigger discussion about it before?

the dubious spellings from the Guidelines are the result of discussion in TraCES.

at least @abausi should have a say; I agree with @CarstenHoffmannMarburg that we must be consistent; if the decision is to adhere to etymology I would then replace all occurences to `mamǝhǝr `

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p><foreign xml:lang="gez">Qǝne</foreign> without type indication in the manuscript. In the first line, it 
+                    refers to <persName ref="PRS3934Ewostate"/>.
+                    It is attributed to <persName ref="PRS14574GabraLeul" cert="high" corresp="SD"/> in the manuscript.</p>
```
DenisNosnitsin1970, line 39: Could you please write out the name?

new/LIT7072QeneMamhera.xml
karljonaskarlsson: Yes, thank you!

new/LIT7072QeneMamhera.xml
karljonaskarlsson: I used the encoding for "ancient correction" according to the guidelines, but I can change to what you suggest if that's better

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="LIT7072QeneMamhera" passive="UppQ1524o"/>
+                </listRelation>
+            </div>
+            <div type="edition" xml:lang="gez">
+                <note>This edition is based on <ref type="mss" corresp="UppQ1524o"/> on fols 40r, 41r. Division into verses is indicated with ፤ .
```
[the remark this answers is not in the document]
karljonaskarlsson: The people at the library took the numbering of the poems to refer to folios, resulting in the rather strange idea that only every second leaf has been foliated… I describe it in the ms description.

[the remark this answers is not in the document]
karljonaskarlsson: "20" is the number of the poem

new/LIT7072QeneMamhera.xml
DenisNosnitsin1970: I would say, we uniformly transliterate (not transcribe) is as "mamhǝr"... The word is in modern languages , that makes the matter complicated.

new/LIT7072QeneMamhera.xml
karljonaskarlsson: Well, you can never know if dabr is "monastery" or "mountain" if it's not clear from the context, right? I think the mention of "dabr" later in the qene is not a name.

new/LIT7072QeneMamhera.xml
```diff
@@ -0,0 +1,82 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7072QeneMamhera" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
```
DenisNosnitsin1970, line 8: ኤዎስጣቴዎስ? It it is አ, I would still make a correction,and encode it as sic in the text below.

new/LIT7072QeneMamhera.xml
DenisNosnitsin1970, line 9: There is no special punctuation marks in the line, and we do not use any, therefore.  Three dots only in translation (as in other qene records).

new/LIT7072QeneMamhera.xml
karljonaskarlsson: You are right! :) Thanks!

DenisNosnitsin1970 (changes requested):

new/LIT7072QeneMamhera.xml
karljonaskarlsson: > If it has eight verses, I would define it as mawaddǝs type according to Guidi 1901.

Ok! I add it.

karljonaskarlsson: Do you mean I should change to a word divider at the end?

new/LIT7072QeneMamhera.xml
karljonaskarlsson: Is this necessary? I thought it is normally enough to refer to the PRS record, or am I wrong?

new/LIT7072QeneMamhera.xml
DenisNosnitsin1970: No, they are fine and mark the verses; i meant two commas in English translation.

new/LIT7072QeneMamhera.xml
DenisNosnitsin1970: Yes, but for the sake of better readability, in a short text, it is word of doing. For instance when I look through it on GitHub, I have to go to the databank for this.

new/LIT7072QeneMamhera.xml
```diff
@@ -5,9 +5,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">">መምህረ፡ ዝንቱ፡ ዓለም፡ አውስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Mamǝhǝra zǝntu ʿālam, ʾAwǝsṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro… (Qǝne of unknown type)</title>
-                <title xml:lang="en" corresp="#t1">The teacher of this world, ʾEwosṭātewos, came to the mountain of Armenia to praise Your Son…</title>
+                <title xml:lang="gez" xml:id="t1">መምህረ፡ ዝንቱ፡ ዓለም፡ <choice><sic>አ</sic><corr resp="JK">ኤ</corr></choice>ዎስጣቴዎስ፡ ደብረ፡ አርማንያ፡ ወልደከ፡ በጽሐ፡ ለዘምሮ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Mamhǝra zǝntu ʿālam, ʾEwosṭātewos, dabra ʾArmānyā waldaka baṣḥa la-zammǝro (Qǝne of unknown type)</title>
```
DenisNosnitsin1970: I meant this: Mamhǝra zǝntu ʿālam ʾEwosṭātewos dabra ʾArmānyā waldaka baṣḥa la-zammǝro, ...

new/LIT7072QeneMamhera.xml
karljonaskarlsson: Ah, ok, I understand. Thanks!

new/LIT7072QeneMamhera.xml
karljonaskarlsson: Ok, I understand

new/LIT7072QeneMamhera.xml
DenisNosnitsin1970: There is a multiplicity of ways one can constract his/her own understanding of verses. It is better to reduce our interpretation exclusively to the translation :)

DenisNosnitsin1970: Please, no mamǝhǝr. It will be very odd.  It is mamhǝr since EAe, the pronunciation is mamhǝr (with a light phonetic schwa before -h) in living languages, and Leslau has it as mamhǝr. Mamǝhǝr in some records may result not from morphological consderations, but from mechanical transliteration.

new/LIT7072QeneMamhera.xml
eu-genia: I am in favour of not writing as it minimizes errors. E. g. if at some point one would say we need to switch to Leslau/EAe transcription I only have to change the spelling once in PRS14574GabraLeul and not in an unlimited nnumber of records pointing to it through `ref`

new/LIT7072QeneMamhera.xml
CarstenHoffmannMarburg: OK.

CarstenHoffmannMarburg (approved):

new/LIT7072QeneMamhera.xml
nafisa-valieva: makes sense!

PR #1214, opened 2024-09-27
Update LIT4274AkkonuB.xml

karljonaskarlsson:
Updated the translation of the title to reflect the contextual meaning of the words

### DIFF
--- a/4001-5000/LIT4274AkkonuB.xml
+++ b/4001-5000/LIT4274AkkonuB.xml
@@ -7,7 +7,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <title xml:lang="gez" xml:id="t1">አኮኑ፡ ብእሲ፡</title>
             <title xml:lang="gez" corresp="#t1" type="normalized">ʾAkko-nu bǝʾsi</title>
             <title xml:lang="en" corresp="#t1">Is it not that a man…?</title>
-            <title xml:lang="en" corresp="#t1">Was he not a man?</title>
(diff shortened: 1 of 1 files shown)


4001-5000/LIT4274AkkonuB.xml
```diff
@@ -6,7 +6,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <titleStmt>
             <title xml:lang="gez" xml:id="t1">አኮኑ፡ ብእሲ፡</title>
             <title xml:lang="gez" corresp="#t1" type="normalized">ʾAkko-nu bǝʾsi</title>
-            <title xml:lang="en" corresp="#t1">Was he not a man?</title>
+            <title xml:lang="en" corresp="#t1">Is it not that a man…?</title>
```
eu-genia, line 9: Can you add the deleted translation back possibly as t2? It is good to preserve labels used historically, and "Was he not a man" is the label used in EMML catalogues, by PEMM, by CANAMEi (https://www.ipocan.it/index.php/it/ms/123-parma-palatina-3835) etc

eu-genia (changes requested):

4001-5000/LIT4274AkkonuB.xml
karljonaskarlsson: Can I have two English translations of t1?

karljonaskarlsson: (I added it as separate t2)

eu-genia (approved):

PR #1215, opened 2024-09-27
2024 09 27 jk qene negusa

karljonaskarlsson:
Another qene from Uppsala, images here: https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0414&pid=alvin-record:280997

### DIFF
--- a/new/LIT7077QeneNegusa.xml
+++ b/new/LIT7077QeneNegusa.xml
@@ -1,0 +1,61 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7077QeneNegusa" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Nǝguśa nagaśt wald ʾama qoma ba-ʿawda tasayṭa salāsā ()</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="JK"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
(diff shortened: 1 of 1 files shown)


nafisa-valieva (approved):
Nice! Looks good and clear for me.

new/LIT7077QeneNegusa.xml
```diff
@@ -0,0 +1,81 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7077QeneNegusa" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ንጉሠ፡ ነገሥት፡ ወልድ፡ አመ፡ ቆመ፡ በዓውደ፡ ተሰይጠ፡ ሰላሳ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Nǝguśa nagaśt wald ʾama qoma ba-ʿawda tasayṭa salāsā (Qǝne of kəbr yəʾəti-type)</title>
```
CarstenHoffmannMarburg, line 9: Even if it is grammatically correct, the manuscript has በዓውደ, so it should be transcribed ba-ʿāwda (not ba-ʿawda).

new/LIT7077QeneNegusa.xml
```diff
@@ -0,0 +1,81 @@
...
+                <title xml:lang="en" corresp="#t1">The king of kings, the Son, when he stood in the courtyard (where) he was sold for thirty (silver coins)…</title>
+                <editor role="generalEditor" key="AB"/>
```
CarstenHoffmannMarburg, line 11: I think, there must be a comma after courtyard and I suggest to change the conjunction after. As far as I know, Jesus was sold, before he was brought to the court:
The king of kings, the Son, when he stood in the courtyard, **for** he was sold for thirty (silver coins)...

CarstenHoffmannMarburg (changes requested):

new/LIT7077QeneNegusa.xml
karljonaskarlsson: You are right! Thank you

new/LIT7077QeneNegusa.xml
karljonaskarlsson: It's a matter of English grammar whether there should be a comma here or not. My feeling is this is a restrictive relative clause and that a comma should therefore not be used, but maybe there are different rules about this.

I'm not sure about translating the asyndetic (relative?) clause as a reason clause, but if you prefer it, I can add it as an option.

CarstenHoffmannMarburg: It is up to you. The conjunction "for" is just a suggestion by myself. If you prefer "where", you can also keep that. An alternative translation is not necessary in my opinion.

The same on the comma. If you think, that a comma is not needed here, you can leave it like this.

CarstenHoffmannMarburg (approved):

PR #1220, opened 2024-09-30
2024 09 30 jk qene wa mota

karljonaskarlsson:
Another qene from the Uppsala collection. This time, the meaning of the first line escapes me almost completely, so I would very much appreciate your input :)

Images here: https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0398&pid=alvin-record:280997

### DIFF
--- a/new/LIT7080Qene.xml
+++ b/new/LIT7080Qene.xml
@@ -42,7 +42,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     However, according to <bibl><ptr target="bm:Guidi1900Qene"/><citedRange>464</citedRange></bibl>, 
                     <foreign xml:lang="gez">śəllāse</foreign>-type <foreign xml:lang="gez">qǝne</foreign>s
                     have six lines. <bibl><ptr target="bm:Habtemichael1998Ufficio"/><citedRange>216</citedRange></bibl>, 
-                    in contrast, who says they have three lines.</p>
+                    in contrast, says they have three lines.</p>
             </abstract>
             <textClass>
                 <keywords>
--- a/new/LIT7080Qene.xml
+++ b/new/LIT7080Qene.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">ወሞት፡ መንፈቀ፡ ስጋነ፡ ለረኃብ፡ ወመንፈቀ፡ ሥጋነ፡ ለመሬት፤</title>
+                <title xml:lang="gez" xml:id="t1">ወሞት፡ መንፈቀ፡ ስጋነ፡ ለረኃብ፡ ወመንፈቀ፡ ሥጋነ፡ ለመ<choice resp="JK"><sic>ሩ</sic><corr>ሬ</corr></choice>ት፤</title>
                 <title xml:lang="gez" corresp="#t1" type="normalized">Wa-mot manfaqa sǝgāna la-raḫāb wa-manfaqa śǝgāna la-maret (Qǝne of the śəllāse-type)</title>
                 <title xml:lang="en" corresp="#t1">And death, half of our flesh for hunger and half of our flesh for the soil…</title>
                 <editor role="generalEditor" key="AB"/>
--- a/new/LIT7080Qene.xml
+++ b/new/LIT7080Qene.xml
@@ -5,7 +5,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title xml:lang="gez" xml:id="t1">Wa-mot manfaqa sǝgāna la-raḫāb wa-manfaqa śǝgāna la-maret (Qəne of tha śəllāse-type)</title>
+                <title xml:lang="gez" xml:id="t1">ወሞት፡ መንፈቀ፡ ስጋነ፡ ለረኃብ፡ ወመንፈቀ፡ ሥጋነ፡ ለመሬት፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Wa-mot manfaqa sǝgāna la-raḫāb wa-manfaqa śǝgāna la-maret (Qǝne of the śəllāse-type)</title>
+                <title xml:lang="en" corresp="#t1">And death, half of our flesh for hunger and half of our flesh for the soil…</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="JK"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -21,7 +23,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </availability>
             </publicationStmt>
             <sourceDesc>
-                <p/>
+                <listWit>
+                    <witness corresp="UppQ1524o"/>
+                </listWit>
             </sourceDesc>
         </fileDesc>
         <encodingDesc>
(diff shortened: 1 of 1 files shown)


new/LIT7080Qene.xml
```diff
@@ -0,0 +1,79 @@
...
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p><foreign xml:lang="gez">Śəllāse</foreign>-type <foreign xml:lang="gez">qǝne</foreign> attributed to 
+                    <persName ref="PRS14574GabraLeul" cert="high" corresp="SD">Gabra Ləʿul</persName> 
+                    in the manuscript. In the manuscript, it appears that the first word of the first line was originally (in the Vorlage?)
+                    intended as a <foreign xml:lang="gez">qǝne</foreign>-type indication.</p>
```
CarstenHoffmannMarburg, line 40: The second sentence is a little complicated and hypothetic. Better: "The <foreign xml:lang="gez">qǝne</foreign>-type is indicated in the beginning of the first line.

new/LIT7080Qene.xml
```diff
@@ -0,0 +1,79 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7080Qene" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ወሞት፡ መንፈቀ፡ ስጋነ፡ ለረኃብ፡ ወመንፈቀ፡ ሥጋነ፡ ለመ<choice resp="JK"><sic>ሩ</sic><corr>ሬ</corr></choice>ት፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Wa-mot manfaqa sǝgāna la-raḫāb wa-manfaqa śǝgāna la-maret (Qǝne of the śəllāse-type)</title>
+                <title xml:lang="en" corresp="#t1">And death, half of our flesh for hunger and half of our flesh for the soil…</title>
```
CarstenHoffmannMarburg, line 10: I think, the translation is Ok. I you are looking for alternatives, here is suggestion (a little more pathetic):
"As for the death, half of our flesh is for hunger, and half of our flesh is for dust"

CarstenHoffmannMarburg (approved):

nafisa-valieva (approved):

DenisNosnitsin1970 (changes requested):

new/LIT7080Qene.xml
CarstenHoffmannMarburg: > Sellase is the type of six verses, and here we have three which corresponds to (za)almakiya. I have no reasonable explanation for Sellase at the beginning, I would just transcribe it as it is and translate as vague as it is.

I this case, it is wrongly attributed to Sallāśe and should be indicated as such in the abstract.

karljonaskarlsson: According to Habtemichael Kidane 1998 ("L'Ufficio divino…"), p. 216, citing an Amharic source, sellase should have three verses.

new/LIT7080Qene.xml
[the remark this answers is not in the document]
karljonaskarlsson: This, for me, is another indication that it's probably Kolmodin's hand. He also confuses ስ and ሰ in a way that suggests that maybe the Vorlage was unclear and he did not know which one would be expected.

new/LIT7080Qene.xml
karljonaskarlsson: I added a note on this, but tell me if you think that this info should rather be placed in the file about "sellase-type qene" (if there is such a file)

new/LIT7080Qene.xml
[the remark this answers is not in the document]
karljonaskarlsson: As you say, the letter in la-maret is very similar to a "ru", so I think it's fair to "correct" it in the edition, as I did (or that's at least what I intended to mark :) )

new/LIT7080Qene.xml
```diff
@@ -0,0 +1,84 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7080Qene" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ወሞት፡ መንፈቀ፡ ስጋነ፡ ለረኃብ፡ ወመንፈቀ፡ ሥጋነ፡ ለመ<choice resp="JK"><sic>ሩ</sic><corr>ሬ</corr></choice>ት፤</title>
```
eu-genia, line 8: no mark up in the title

eu-genia (changes requested):

new/LIT7080Qene.xml
CarstenHoffmannMarburg: Yes, I agree. If there is contradictory information about what is Śəllāse type in qəne, the record should be reworked. We need to discuss, which standard should be used in Bm.

new/LIT7080Qene.xml
```diff
@@ -0,0 +1,84 @@
...
+            <abstract>
+                <p><foreign xml:lang="gez">Qǝne</foreign> with three lines attributed to 
+                    <persName ref="PRS14574GabraLeul" cert="high" corresp="SD">Gabra Ləʿul</persName> 
+                    in the manuscript. It appears that the first word of the first line was originally (in the Vorlage?)
+                    intended as a <foreign xml:lang="gez">qǝne</foreign>-type indication, identifying this as a 
+                    <foreign xml:lang="gez">śəllāse</foreign>-type <foreign xml:lang="gez">qǝne</foreign>.
+                    However, according to <bibl><ptr target="bm:Guidi1900Qene"/><citedRange>464</citedRange></bibl>, 
+                    <foreign xml:lang="gez">śəllāse</foreign>-type <foreign xml:lang="gez">qǝne</foreign>s
+                    have six lines. <bibl><ptr target="bm:Habtemichael1998Ufficio"/><citedRange>216</citedRange></bibl>, 
+                    in contrast, who says they have three lines.</p>
```
CarstenHoffmannMarburg, line 45: The last one is not a proper sentence. You can style it either: ".... while Habtemichael says they have three lines." or "Habtemichael on the contrary says that they have three lines" or you turn the order of the arguments:

`It appears that the first word of the first line was intended as an indication of the <foreign xml:lang="gez">qǝne</foreign>-type, identifying this as a <foreign xml:lang="gez">śəllāse</foreign>-type <foreign xml:lang="gez">qǝne</foreign> according to <bibl><ptr target="bm:Habtemichael1998Ufficio"/><citedRange>216</citedRange></bibl>, who defines <foreign xml:lang="gez">śəllāse</foreign>-type to have three lines contrary to <bibl><ptr target="bm:Guidi1900Qene"/><citedRange>464</citedRange></bibl>, who describes it to have six lines. `

CarstenHoffmannMarburg (approved):

new/LIT7080Qene.xml
karljonaskarlsson: Thanks!

CarstenHoffmannMarburg (approved):

eu-genia (approved):

karljonaskarlsson:
@DenisNosnitsin1970 , can this be approved?

new/LIT7080Qene.xml
DenisNosnitsin1970: This is a difficult case. For the moment, we stick to the system as exposed by Guidi in his "Qene o inni abissini" where he explains Sellase as a type of 6 rhymed lines. There are some other authors who might have deviations, but I suggest we better stick to this one, otherwise we will get confused. Habtemikael sometimes says things that deviate from what seems to be a common knowledge (he is from Eritrea, by the way; he is not a qene master). In EAE he says Sellase is composed of six "bets" (= three lines; vey logical as it is triple), but elsewhere I did see Sellase made of six lines, identified like this. Before Habtemikael, we have a few other authors who in my opinion are more trustful. "Sellase" at the beginning - as I told, I would leave it as vague as it is. You are not obliged to see an indication to qene-type in it. Another issue is that in such form the line does not make sense, seemingly ("Trinity and death..."???), but I had a few cases when a qene made absolutely no sense in the beginning, and this changed after a while (and many will remain untransparent).  However, I think there is no point in endless discussion here. I just approve, the rest is in your hands.

DenisNosnitsin1970 (approved):

PR #1222, opened 2024-10-01
Create LIT7081NamesGod

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2618.

If there is a decision about keywords for Magic (cf. discussions in https://github.com/BetaMasaheft/Documentation/issues/2627 and in https://github.com/BetaMasaheft/Authority-Files/pull/33), I would propose AsmatGod i.e. "Names of God" as keyword for this type of a Magic prayer (or rather a Christian text, that was used for Magic purpose).

### DIFF
--- a/new/LIT7081NamesGod.xml
+++ b/new/LIT7081NamesGod.xml
@@ -55,6 +55,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="LIT7081NamesGod" passive="BLorient11602"/>
                 </listRelation>
+                <listBibl type="secondary">
+                    <bibl>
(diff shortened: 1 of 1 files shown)


nafisa-valieva:
Thank you! I have no objections!

eu-genia (approved):

CarstenHoffmannMarburg:
@eu-genia : Two weeks after merging, this file cannot be found in the app. What happened?

eu-genia:
no idea, some bug during the record creation

again, an explanation: records should be visible online NOT from the moment of the merge (merge = synchronizing github and app) BUT from the moment they are created in the app (so BEFORE they are present on any branch or any pull request), so as soon as you create an ID you can access it online (but not on github or locally). so two minutes or two weeks after merging does not play a role.

should be there now

PR #1227, opened 2024-10-04
Create LIT7086QeneBoSadqan.xml

karljonaskarlsson:
Another qene from Uppsala, this time from: https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0358&pid=alvin-record:280997

### DIFF
--- a/new/LIT7086QeneBoSadqan.xml
+++ b/new/LIT7086QeneBoSadqan.xml
@@ -1,0 +1,90 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7086QeneBoSadqan" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ቦ፡ ጻድቃን፡ ውሉ<del rend="strikethrough">ድ</del><add place="above">ደ</add>፡ 
+                    ቤትከ፡ እለ፡ መነኑ፡ ዓለ<del rend="strikethrough">ም</del><add place="above">መ</add>፡ ወዕለ፡ ኤሉ፡ አድባረ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Bo ṣādqān wǝluda betǝka ʾǝlla mannanu ʿālama wa-ʿǝlla ʾelu ʾadbāra (Qǝne possibly of the mawaddǝs-type)</title>
+                <title xml:lang="en" corresp="#t1">There are righteous ones, the children of Your house, who despised the world and who roamed the mountains…</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="JK"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <listWit>
+                    <witness corresp="UppQ1524o"/>
+                </listWit>
+            </sourceDesc>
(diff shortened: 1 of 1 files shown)


new/LIT7086QeneBoSadqan.xml
```diff
@@ -0,0 +1,90 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7086QeneBoSadqan" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ቦ፡ ጻድቃን፡ ውሉ<del rend="strikethrough">ድ</del><add place="above">ደ</add>፡
```
CarstenHoffmannMarburg, line 8: I recommend to use strikethrough and add only in body, but not in the title.

new/LIT7086QeneBoSadqan.xml
```diff
@@ -0,0 +1,90 @@
...
+                <ab>
+                    <l n="1">ቦ፡ ጻድቃን፡ ውሉ<del rend="strikethrough">ድ</del><add place="above">ደ</add>፡ 
+                        ቤትከ፡ እለ፡ መነኑ፡ ዓለ<del rend="strikethrough">ም</del><add place="above">መ</add>፡ ወዕለ፡ ኤሉ፡ አድባረ፤</l>
+                    <l n="2">እንዘ፡ ይሐስሱ፡ ዮም፡ ሰማያዊተ፡ ሀገ<del rend="strikethrough">ር</del><add place="above">ረ</add>፤</l>
+                    <l n="3">እስከነ፡ ቀነዩ፡ ሥጋሆሙ፡ ገብረ፤</l>
+                    <l n="4">ወልቡናሆሙ፡ ማዕምረ፤</l>
+                    <l n="5">ለተቀንዎ፡ አምላክ፡ ዘኢያህጸርከ፡ አእጋረ፤</l><pb n="19r"/>
+                    <l n="6">ንህነሂ፡ እለ፡ ጸሐፍነ፡ ሐኬተ፡ ልቡና፡ <choice resp="JK"><sic>ጠ</sic><corr>ጦ</corr></choice>ማረ፤</l>
+                    <l n="7">ወአምረርነ፡ ቃላቲከ፡ ሐሊበ፡ ወመአረ፤</l>
+                    <l n="8">እንዘ፡ ተድላ፡ ዓለም፡ ንሐስስ፡ ወንትሜነዶ፡ ክብረ፤</l>
```
CarstenHoffmannMarburg, line 83: The ዓ in ዓለም፡ looks very much like ፃ. You may decide, if you want to keep it like this or mark (your) correction to ዓለም፡.

CarstenHoffmannMarburg (approved):

new/LIT7086QeneBoSadqan.xml
CarstenHoffmannMarburg: OK. Perhaps the thin stroke inside ዓ was unintentionally.

karljonaskarlsson: I would say that it probably is an error, but I don't know how to markup this kind of ambiguous characters

CarstenHoffmannMarburg: What you can do, is to explain the case explicitly in the abstract, but I do not think, that this case is worth being discussed in the abstract.

karljonaskarlsson: Yes, I agree that it goes a bit too much into detail. Maybe I can remark on this and other similar phenomena in the hand description in the manuscript record (if I remember)

nafisa-valieva (approved):

eu-genia (changes requested):
No mark up in title

eu-genia (approved):

new/LIT7086QeneBoSadqan.xml
```diff
@@ -0,0 +1,90 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7086QeneBoSadqan" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ቦ፡ ጻድቃን፡ ውሉደ፡ 
+                    ቤትከ፡ እለ፡ መነኑ፡ ዓለመ፡ ወዕለ፡ ኤሉ፡ አድባረ፤</title>
```
DenisNosnitsin1970, line 9: Perhaps it is better to finish t1 with ፡ since this is the title of the poem. The proper punctuation will be provided in "Text".

karljonaskarlsson: I have been using the "line-final" punctuation marks regularly in "my" qene records. I could change them all, if this is better. Do you agree, @eu-genia ?

PR #1232, opened 2024-10-04
Create LIT7091QeneAyte.xml

karljonaskarlsson:
Another qene from Uppsala, with images here: https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0318&pid=alvin-record:280997

### DIFF
--- a/new/LIT7091QeneAyte.xml
+++ b/new/LIT7091QeneAyte.xml
@@ -1,0 +1,89 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7091QeneAyte" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">አይ<corr>ቴ</corr>፡ ናስምክ፡ ውሉድ፡ አመጻ፡ ዘተመዝበረ፡ ልብነ፡ በግብረ፡ አመጻ፡ ነፋስ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">ʾAyte nāsmǝk wǝlud ʾamaṣā za-tamazbara lǝbbǝna ba-gǝbra 
+                    ʾamaṣā nafās (Qǝne of the mawaddǝs-type)</title>
+                <title xml:lang="en" corresp="#t1">Where shall we lay, o children, the iniquity that has destroyed our hearts 
+                    through the work(s) of iniquity, wind…</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="JK"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <listWit>
+                    <witness corresp="UppQ1524o"/>
+                </listWit>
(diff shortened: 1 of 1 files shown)


new/LIT7091QeneAyte.xml
```diff
@@ -0,0 +1,89 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7091QeneAyte" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">አይ<corr>ቴ</corr>፡ ናስምክ፡ ውሉድ፡ አመጻ፡ ዘተመዝበረ፡ ልብነ፡ በግብረ፡ አመጻ፡ ነፋስ፤</title>
```
eu-genia, line 8: again, I do not think editorial mark up or punctuation belong to the title of the record

new/LIT7091QeneAyte.xml
```diff
@@ -0,0 +1,89 @@
...
+                <note>This edition is based on <ref type="mss" corresp="UppQ1524o"/> on fols 8r, 9r.
+                    Division into verses is indicated with ፤ .
+                    The end of the poem is indicated with ። .</note>
+            <ab>
+                <l n="1">አይ<corr>ቴ</corr>፡ ናስምክ፡ ውሉድ፡ አመጻ፡ ዘተመዝበረ፡ ልብነ፡ በግብረ፡ አመጻ፡ ነፋስ፤</l>
+                <l n="2"><sic>እምትዕይቱ፡</sic> ሰማየ፡ አመ፡ ይመጽእ፡ ንጉሥ፤</l>
+                <l n="3">በንጥረ፡ መባርቅት፡ ረዋዲ፡ ፈረስ፤</l>
+                <l n="4">ወከመ፡ አማጽያነ፡ ያጠፍዕ፡ በነገረ፡ ተስናን፡ ወባዕስ፤</l>
+                <l n="5">ህዳጠሒ፡ ንጸወን፡ በደብረ፡ ጽወን፡ ሐዳስ፤</l>
+                <l n="6">ደብረ፡ ጽዮን<choice resp="JK"><sic>ስ</sic><corr>ሰ</corr></choice>፡ እስመ፡
```
eu-genia, line 80: I think resp should go with corr

eu-genia (changes requested):

new/LIT7091QeneAyte.xml
```diff
@@ -0,0 +1,89 @@
...
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7091QeneAyte" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">አይ<corr>ቴ</corr>፡ ናስምክ፡ ውሉድ፡ አመጻ፡ ዘተመዝበረ፡ ልብነ፡ በግብረ፡ አመጻ፡ ነፋስ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">ʾAyte nāsmǝk wǝlud ʾamaṣā za-tamazbara lǝbbǝna ba-gǝbra 
+                    ʾamaṣā nafās (Qǝne of the mawaddǝs-type)</title>
+                <title xml:lang="en" corresp="#t1">Where shall we lay, o children, the iniquity that has destroyed our hearts 
+                    through the work(s) of iniquity, wind…</title>
```
CarstenHoffmannMarburg, line 12: "through the iniquity of the spirit"

CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg (approved):

new/LIT7091QeneAyte.xml
karljonaskarlsson: I think they rather compare the works of iniquity with wind

DenisNosnitsin1970: Yes, you should understand it as "the works/deeds of iniquity (which is) wind". Wind may be desctructive and takes away, esp. light things (in this case, we should also consider that "heart" means "intellect"). Maybe th translation can be a little different: Where shall we lay, children, the iniquity having destroyed our hearts though ..." But not necessarily. Nice qene. Maybe, consider also "lean upon" instead of "lay".

DenisNosnitsin1970 (changes requested):

DenisNosnitsin1970 (approved):

new/LIT7091QeneAyte.xml
DenisNosnitsin1970: That is true, it is a title.

new/LIT7091QeneAyte.xml
DenisNosnitsin1970: I would leave it as it is, with `<sic/>`

eu-genia (approved):

PR #1241, opened 2024-10-14
2024 10 11 jk sana som

karljonaskarlsson:
Another composition from the notebook of qenes in Uppsala (https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0302&pid=alvin-record:280997). It has an explicit qene-type ("Guba'e Qana"), but I am not able to discern any rhyme. If you note something, please let me know.

### DIFF
--- a/new/LIT7097SanaSom.xml
+++ b/new/LIT7097SanaSom.xml
@@ -71,6 +71,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <relation name="lawd:hasAttestation" active="LIT7097SanaSom" passive="UppQ1524o"/>
                     <relation name="ecrm:P129_is_about" active="LIT7097SanaSom" passive="PRS3430DawitII"/>
                     <relation name="saws:contains" active="LIT7097SanaSom" passive="LIT7102QeneYom"/>
+                    <relation name="saws:contains" active="LIT7097SanaSom" passive="LIT7114QeneSana"/>
                 </listRelation>
             </div>
             <div type="edition" xml:lang="gez">
--- a/new/LIT7097SanaSom.xml
+++ b/new/LIT7097SanaSom.xml
@@ -35,14 +35,21 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <creation/>
             <abstract>
                     <p>Poetic composition, defined in the manuscript as a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>.
-                        However, no rhymes can be discerned, as one would expect in a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type
-                        <foreign xml:lang="gez">qəne</foreign>. 
-                        The composition has no author attribution in the manuscript.</p>
+                        The composition has no author attribution in the manuscript. In fact, the work appears to be composed of two
+                        different <foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type <foreign xml:lang="gez">qəne</foreign>s
+                        and one spurious line. The first <foreign xml:lang="gez">qəne</foreign> is made up of lines 5–7, which
+                        should be reanalyzed as two lines by uniting lines 1–5 and lines 6–7, respectively. The second 
+                        <foreign xml:lang="gez">qəne</foreign>, which also has a separate circulation in a printed book
+                        (<bibl><ptr target="bm:Afawarq1984Yageezqeneyat"></ptr><citedRange unit="page">230-231, no. 246</citedRange></bibl>), 
+                        is made up of lines 9–12, which shoyld be reanalyzed as two lines by coalescing lines 10–12 into one line.
+                        Line 8 is spurious.</p>
             </abstract>
             <textClass>
                 <keywords>
                     <term key="ChristianLiterature"/>
                     <term key="Poetry"/>
+                    <term key="Qene"/>
+                    <term key="GubaeQana"/>
                 </keywords>
             </textClass>
             <langUsage>
@@ -59,6 +66,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <div type="bibliography">
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="LIT7097SanaSom" passive="UppQ1524o"/>
+                    <relation name="saws:contains" active="#LIT7097SanaSom" passive="LIT7102QeneYom"/>
                 </listRelation>
             </div>
             <div type="edition" xml:lang="gez">
--- a/new/LIT7097SanaSom.xml
+++ b/new/LIT7097SanaSom.xml
@@ -45,10 +45,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <term key="Poetry"/>
                 </keywords>
             </textClass>
-            <t:langUsage xmlns:t="http://www.tei-c.org/ns/1.0">
+            <langUsage>
                 <language ident="en">English</language>
                 <language ident="gez">Gǝʿǝz</language>
-            </t:langUsage>
+            </langUsage>
         </profileDesc>
         <revisionDesc>
             <change who="JK" when="2024-10-11">Created entity</change>
--- a/new/LIT7097SanaSom.xml
+++ b/new/LIT7097SanaSom.xml
@@ -6,8 +6,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">ጸንአ፡ ጾም፡ ነግረ፡ ኢሳይያ፡ አብ፤</title>
-                <title xml:lang="gez" corresp="#t1" type="normalized">Ṣanʾa ṣom nagra ʾIsāyǝyā ʾab (Qǝne of the mawaddǝs-type)</title>
-                <title xml:lang="en" corresp="#t1">The fast grew strong, said father ʾIsāyǝyā</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Ṣanʾa ṣom nagra ʾIsāyǝyā ʾab (Poetic composition)</title>
+                <title xml:lang="en" corresp="#t1">The fast grew strong, said father ʾIsāyǝyā…</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="JK"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -34,9 +34,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                    <p>Poetic composition, defined in the manuscript as a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>,
-                        but no rhymes can be discerned. 
-                        It has no attribution in the manuscript.</p>
+                    <p>Poetic composition, defined in the manuscript as a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>.
+                        However, no rhymes can be discerned, as one would expect in a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type
+                        <foreign xml:lang="gez">qəne</foreign>. 
+                        The composition has no author attribution in the manuscript.</p>
             </abstract>
             <textClass>
                 <keywords>
@@ -63,16 +64,18 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <div type="edition" xml:lang="gez">
                 <note>This edition is based on <ref type="mss" corresp="UppQ1524o"/> on fols 4r, 5r.
                     Division into verses is indicated with ፤ .
-                    The end of the poem, as well as some poem-internal verse divisions, are indicated with ። .</note>
+                    The end of the poem, as well as some poem-internal verse divisions, are indicated with ። .
+                    Line 5 is indented.
+                </note>
                 <ab>
                     <l n="1">ጸንአ፡ ጾም፡ <sic resp="JK">ነግረ፡</sic> ኢሳይያ፡ አብ፤</l>
                     <l n="2">አምጣነ፡ ባቲ፡ ወአሉ፤</l>
                     <l n="3">እስከ፡ ፍና፤</l>
                     <l n="4">ሰርክ፡ ሕዝብ፤</l>
                     <l n="5">መፍትው።</l>
-                    <l n="6">ይትፌሳሕ፡ ቀኖና፡ ሥጋ፤</l><pb n="9r"/>
+                    <l n="6">ይትፌሳሕ፡ ቀኖና፡ ሥጋ፤</l>
                     <l n="7">በጽርሁ፡ ቀርነ፡ እንስሳ፥ እልው።</l><pb n="5r"/>
-                    <l n="8">እስመ፡ ተኃንቀ፡ ጉርኤሁ፤</l>
+                    <l n="8">እስመ፡ ተኃንቀ፡ ጉርኤሁ።</l>
                     <l n="9">ዮም፡ ዕለት፡ ብዙሕ፡ ትፍስሕት፤</l>
                     <l n="10">እስመ፡ በደብረ፡ ታቦር፤</l>
                     <l n="11">ቆመ፡ ፈጣሪ፤</l>
--- a/new/LIT7097SanaSom.xml
+++ b/new/LIT7097SanaSom.xml
@@ -7,7 +7,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">ጸንአ፡ ጾም፡ ነግረ፡ ኢሳይያ፡ አብ፤</title>
                 <title xml:lang="gez" corresp="#t1" type="normalized">Ṣanʾa ṣom nagra ʾIsāyǝyā ʾab (Poetic composition)</title>
-                <title xml:lang="en" corresp="#t1">The fast grew strong, said father ʾIsāyǝyā…</title>
+                <title xml:lang="en" corresp="#t1">The fasting grew strong, said Father ʾIsāyǝyā…</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="JK"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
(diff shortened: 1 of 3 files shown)


karljonaskarlsson:
(If we decide to use word-dividers at the end of t1, I will change it in all "my" files at the same time.)

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,87 @@
...
+            <creation/>
+            <abstract>
+                    <p>Poetic composition, defined in the manuscript as a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>.
+                        However, no rhymes can be discerned, as one would expect in a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type
+                        <foreign xml:lang="gez">qəne</foreign>. 
+                        The composition has no author attribution in the manuscript.</p>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
```
eu-genia, line 44: You think it is not justified to add Qene at least as keyword here?

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,87 @@
...
+                    <term key="Poetry"/>
+                </keywords>
+            </textClass>
+            <t:langUsage xmlns:t="http://www.tei-c.org/ns/1.0">
```
eu-genia, line 48: remove t: and xmlns (sometimes the app inserts this but better to clean)

eu-genia (changes requested):

new/LIT7097SanaSom.xml
karljonaskarlsson: I'm not sure. What do you say, @DenisNosnitsin1970 and @CarstenHoffmannMarburg ?

CarstenHoffmannMarburg: > It has an explicit qene-type ("Guba'e Qana"), but I am not able to discern any rhyme. If you note something, please let me know.

It has a crossed rhyme ending in ብ abd ሉ/ልው. Important for the definition of Guba'e Qana is the rhyme in አሉ/እልው, because Guba'e Qana has only two verses according to Guidi.
I think, it is legitimate to add the keywords Qene and Guba'e Qana and note the extraordinary rhyme.

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,87 @@
...
+                    <l n="3">እስከ፡ ፍና፤</l>
+                    <l n="4">ሰርክ፡ ሕዝብ፤</l>
+                    <l n="5">መፍትው።</l>
+                    <l n="6">ይትፌሳሕ፡ ቀኖና፡ ሥጋ፤</l>
+                    <l n="7">በጽርሁ፡ ቀርነ፡ እንስሳ፥ እልው።</l><pb n="5r"/>
+                    <l n="8">እስመ፡ ተኃንቀ፡ ጉርኤሁ።</l>
+                    <l n="9">ዮም፡ ዕለት፡ ብዙሕ፡ ትፍስሕት፤</l>
+                    <l n="10">እስመ፡ በደብረ፡ ታቦር፤</l>
+                    <l n="11">ቆመ፡ ፈጣሪ፤</l>
+                    <l n="12">ወበደብረ፡ ብርሃን፡ ዳዊት።</l>
```
CarstenHoffmannMarburg, line 82: Because it is Guba'e Qana, it should be given in two verses (`<lb>`) ending in the rhyme አሉ/እልው.

new/LIT7097SanaSom.xml
karljonaskarlsson: Are you suggesting that this is in fact two different two-line qenes?

new/LIT7097SanaSom.xml
karljonaskarlsson: Very interesting! Ok, then I divide it up into two, ok?

I can't find a LIT ID for the qene you mention, Denis, so I will go ahead and create one for it (with the Uppsala manuscript as a manuscript witness and the printed book as an edition)

DenisNosnitsin1970: It is not in Bm yet, here is a portion of the book, see no. 246 (the excerpt has no title, I have the book at home and will check). 
[library63a8.pdf](https://github.com/user-attachments/files/17362678/library63a8.pdf)

nafisa-valieva (approved):

new/LIT7097SanaSom.xml
karljonaskarlsson: Thank you!

new/LIT7097SanaSom.xml
```diff
@@ -59,6 +66,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <div type="bibliography">
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="LIT7097SanaSom" passive="UppQ1524o"/>
+                    <relation name="saws:contains" active="#LIT7097SanaSom" passive="LIT7102QeneYom"/>
```
eu-genia, line 69: remove `#`

new/LIT7097SanaSom.xml
```diff
@@ -59,6 +66,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                 </listRelation>
             </div>
             <div type="edition" xml:lang="gez">
```
eu-genia, line 72: I would add `div `for the parts in the edition, and a `corresp `to the lines corresponding to 7102

new/LIT7102QeneYom.xml
```diff
@@ -0,0 +1,80 @@
...
+        <revisionDesc>
+            <change who="JK" when="2024-10-15">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="saws:isAttributedToAuthor" active="LIT7102QeneYom" passive="PRS14578KeflaYohannes"/>
+                    <relation name="ecrm:P129_is_about" active="#LIT7102QeneYom" passive="PRS3430DawitII"/>
```
eu-genia, line 63: remove `#`

new/LIT7097SanaSom.xml
eu-genia: not isAbout relation here as well?

eu-genia (changes requested):

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,87 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7097SanaSom" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ጸንአ፡ ጾም፡ ነግረ፡ ኢሳይያ፡ አብ፤</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Ṣanʾa ṣom nagra ʾIsāyǝyā ʾab (Poetic composition)</title>
```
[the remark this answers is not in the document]
karljonaskarlsson: But this is not one qene of Gubae Qana type, as I explain in the abstract. In the layout of the Uppsala manuscript, this is treated as one text, but as I think we have discovered (and as I try to explain in the abstract), it is made up of several parts, two of which are qene of the Guba'e Qana type. But I don't think we can say that this composite composition is.

new/LIT7102QeneYom.xml
```diff
@@ -0,0 +1,80 @@
...
+            </div>
+            <div type="edition" xml:lang="gez">
+                <note>This edition is based on 
+                    <bibl><ptr target="bm:Afawarq1984Yageezqeneyat"></ptr><citedRange unit="page">230</citedRange></bibl>.
+                    Division into verses is indicated with ፡ , whereas words are separated by spaces without word dividers.
+                    The end of the poem is indicated with ። .
+                </note>
+                <ab>
+                    <l n="1">ዮም ዕለተ ብዙኅ ትፍሥሕት፡</l>
+                    <l n="2">እስመ በደብረ ታቦር ቆመ ፈጣሪ ወበደብረ ብርሃን ዳዊት።</l>
```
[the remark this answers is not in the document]
karljonaskarlsson: But in the manuscript, this text is not transmitted as an individual text, but only as part of a larger whole. That's why I chose to base the edition on the printed edition. But in any case, I can add word dividers if it enhances the usability of the text

new/LIT7097SanaSom.xml
karljonaskarlsson: I'm not sure I understood precisely how you mean, so please check if what I did works

new/LIT7097SanaSom.xml
karljonaskarlsson: Based on what Denis commented, I would analyse what we have in the Uppsala manuscript as _one_ text, because it follows the layout for individual texts in the manuscript and has one number in the numbering of the manuscript, etc. But this text is not strictly a qene, because it breaks the rules. Instead, I would say that it is composed of two qenes, one of which is attested elsewhere, as noted by Denis, and therefore should have its own LIT ID (= LIT7102). This is what I tried to convey in the abstract :)

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,103 @@
...
+                    Division into verses is indicated with ፤ .
+                    The end of the poem, as well as some poem-internal verse divisions, are indicated with ። .
+                    Verse 5 is indented.
+                </note>
+                <div type="textpart" xml:id="p1">
+                    <l n="1">ጸንአ፡ ጾም፡ <sic resp="JK">ነግረ፡</sic> ኢሳይያ፡ አብ፤</l>
+                    <l n="2">አምጣነ፡ ባቲ፡ ወአሉ፤</l>
+                    <l n="3">እስከ፡ ፍና፤</l>
+                    <l n="4">ሰርክ፡ ሕዝብ፤</l>
+                    <l n="5">መፍትው።</l>
```
eu-genia, line 87: I understand it that p1 is analyzed by @DenisNosnitsin1970 as containing two lines (currently 1-5 and 6-7); the punctuation mark ፤ is just a semicolon and not a line divider

eu-genia: so we have 
`poem1 `with (it is OK to use whichever xml:id but usually we think `p` is part in manuscripts)
line 1 (now 1-5)
line 2 (now 6-7)

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,103 @@
...
+                    <l n="6">ይትፌሳሕ፡ ቀኖና፡ ሥጋ፤</l>
+                    <l n="7">በጽርሁ፡ ቀርነ፡ እንስሳ፥ እልው።</l><pb n="5r"/>
+                </div>
+                <div type="textpart" xml:id="p2">
+                    <l n="8">እስመ፡ ተኃንቀ፡ ጉርኤሁ።</l>
```
eu-genia, line 92: this is then line 3

new/LIT7097SanaSom.xml
eu-genia: And `poem3 `(`corresp `is correct, thank you) contains I guess
line 4 (formerly 9)
line 5 (formerly 10-12)

eu-genia (changes requested):

new/LIT7102QeneYom.xml
```diff
@@ -0,0 +1,85 @@
...
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p><foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type <foreign xml:lang="gez">qəne</foreign>. It has been
+                    published in <bibl><ptr target="bm:Afawarq1984Yageezqeneyat"></ptr><citedRange unit="page">230-231, no. 246</citedRange></bibl>,
```
eu-genia, line 36: `<citedRange unit="page">320-321</citedRange>, <citedRange unit="number">246</citedRange>
`

new/LIT7102QeneYom.xml
```diff
@@ -0,0 +1,85 @@
...
+                    where it is attributed to <persName ref="PRS14578KeflaYohannes" role="author"/>. It also appears as part of
```
eu-genia, line 37: According to the edition, it is chanted in `<ref type="authFile" corresp="EzelMode">ʿǝzǝl mode</ref>`

new/LIT7102QeneYom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="saws:isAttributedToAuthor" active="LIT7102QeneYom" passive="PRS14578KeflaYohannes"/>
+                    <relation name="ecrm:P129_is_about" active="LIT7102QeneYom" passive="PRS3430DawitII"/>
+                    <relation name="saws:formsPartOf" active="LIT7102QeneYom" passive="LIT7097SanaSom"/>
+                </listRelation>
+            </div>
+            <div type="edition" xml:lang="gez">
+                <note>This edition is based on 
+                    <bibl><ptr target="bm:Afawarq1984Yageezqeneyat"></ptr><citedRange unit="page">230</citedRange></bibl>.
```
eu-genia, line 69: I guess the page is 320 (see also above)

eu-genia (changes requested):

new/LIT7097SanaSom.xml
```diff
@@ -0,0 +1,95 @@
...
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                    <p>Poetic composition, defined in the manuscript as a <foreign xml:lang="gez">Gubāʾa Qānā</foreign>.
+                        The composition has no author attribution in the manuscript. In fact, the work appears to be composed of two
+                        different <foreign xml:lang="gez">Gubāʾa Qānā</foreign>-type <foreign xml:lang="gez">qəne</foreign>s
+                        and one spurious line. The first <foreign xml:lang="gez">qəne</foreign> is made up of lines 5–7, which
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: I understand "lines" as written lines according to the layout description in our [Guidelines](https://betamasaheft.eu/Guidelines/?q=line&id=layoutDescPage). In contrast, I understand "verses" as Qǝne verses in this context.

new/LIT7097SanaSom.xml
karljonaskarlsson: You suggest to give the edition based on our interpretation of the data, rather than what is found in the manuscript. Granted that we reanalyse the entire composition as strange mix of two qenes plus one extra line, I think it's better to keep the line analysis of the manuscript and describe our interpretation in words (as I did). Is this ok?

new/LIT7102QeneYom.xml
karljonaskarlsson: (This is not how I have done it in catalogue descriptions, however)

new/LIT7097SanaSom.xml
eu-genia: I suggest that the text edition (in LIT records) is about text, and transcription (in Manuscripts) follows the manuscripts; in this case Denis's view is also confirmed by the printed edition of one of the parts

karljonaskarlsson: As I understand it, the manuscript attests to a composite text to which it is the only witness. It may be that this composite text probably originated in a misunderstanding of the sources (= the copyist merged at least two different qene into one text), but now it is nevertheless _one_ text (according to layout in the manuscript, numbering, etc.). This text contains another text, which is the qene identified by Denis and found in the printed book. But the rest of the composite text has no attestation outside of this context, as far as we are aware.

LIT7097SanaSom is for the composite text. This text has no unifying rhyme and I don't think it makes sense to try to trace its history in the edition. I would say it's better to present the text as it is transmitted.

new/LIT7097SanaSom.xml
DenisNosnitsin1970: Still I would divide the poems, because in work records we treat the works irrespective of the way they are shaped in the manuscripts. We assume that both poems were created as separate compositions and only by chance were copied together, with an obscure line in between (unless it is an incredible case of "double qene"). If the structure of the written lines do not fit the structure prescribed for the qene type, this is the phenomenon of the text arrangement in the manuscript. In other words, the qene ጸንአ፡ ጾም፡ ... fits guba'e qana type, but it's two long verses are simply not copied in two lines, by whatever reasons. But for us it is absolutely justified to present the poem in its work ID as composed of two lines (መፍትው - እልው), and (provisionally) assigned a type to it. If you want to demonstrate its shape in the manuscript, you transcribe the poem in the same structure in the record for manuscript (then, division in verses/lines and written lines will be as you indicated here). In a similar way, we do not reproduce rubrications in work ID - all this pertains to shaping the text in concrete witnesses and should be treated in manuscript records. Also, the fact that one of the qenes is recorded in a publication is another argument for separating them.

DenisNosnitsin1970 (changes requested):

new/LIT7097SanaSom.xml
karljonaskarlsson: As I understand work records, we want to record works that are circulating in the Ethiopian manuscript culture. I agree with your analysis that what we find in the Uppsala manuscript must originally have been two guba'e qana qenes plus one spurious line – but the manuscript has what it has, which is clearly something analysed (probably wrongly! :) ) as one work.

I have a suggestion: would it be OK if I create a separate ID for the first, hypothetical qene (the two lines መፍትው - እልው), parallel to the one I created for Yom 3elätät. And then I still keep one "general record" for the entire text as it is attested in the Uppsala manuscript, i.e. the two qenes plus the spurious line. Otherwise, I would need to create a separate LIT ID for the spurious line as well, and it feels like that is taking it too far (and it also departs farther and farther away from how we define a work, doesn't it?)

DenisNosnitsin1970: Creating ID for the first qene and a separate ID for the entire block is ok, but you are not obliged to create an ID for one line, because it is not a work. We do not know exactly how the passage was understood by the writer. Even though, the way it is copied is the phenomenon of the transmission in concrete witnesses.  If you work in manuscripts, you will find a lot of "noise" - separate lines and small text bits, which are not works (they may be parts of some works, but to trace back everything is not realistic).

karljonaskarlsson:
@eu-genia @DenisNosnitsin1970 , are there still matters you would like me to attend to before this can be approved?

DenisNosnitsin1970 (approved):

eu-genia (approved):

PR #1250, opened 2024-10-17
2024 10 14 jk qene leyon

karljonaskarlsson:
Another qene from Uppsala (although this is one of the last – maybe to the relief of some!): https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0471&pid=alvin-record:280997

It is attributed to a "Twoldo" in the manuscript, but I have not been able to identify him further (and it seems like this is only half of his name). Should I have created a PRS record for him nonetheless? (I have not done it so far).

### DIFF
--- a/new/LIT7100QeneLeyon.xml
+++ b/new/LIT7100QeneLeyon.xml
@@ -64,10 +64,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <div type="bibliography">
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="LIT7100QeneLeyon" passive="UppQ1524o"/>
-                    <relation name="ecrm:P129_is_about"  active="LIT7072QeneMamhera" passive="PRS6255LeoI"/>
-                    <relation name="ecrm:P129_is_about"  active="LIT7072QeneMamhera" passive="PRS3919Eutychiu"/>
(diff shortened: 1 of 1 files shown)


eu-genia (approved):

CarstenHoffmannMarburg (approved):

nafisa-valieva (approved):

new/LIT7100QeneLeyon.xml
```diff
@@ -0,0 +1,86 @@
...
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p><foreign xml:lang="gez">Wāzemā</foreign>-type <foreign xml:lang="gez">qǝne</foreign> 
+                    attributed in the manuscript to a "Twoldo" who cannot be further identified.
```
DenisNosnitsin1970, line 38: I think we add our interpretation of "Twoldo" as "Tawalda", which is quite secure, I think.

DenisNosnitsin1970: Well, I think creating an ID for "Twoldo"/"Tawalda" him would not be wrong.

karljonaskarlsson: Ok, I will let create it

PR #1251, opened 2024-10-17
Update LIT1650Homily.xml

MarcinKrawczuk:
I've noticed that my changes are not visible in the app, so I am re-merging, no new changes.


eu-genia:
If something is there on the master branch in GitHub but you do not see it in the app just let me know

PR #1253, opened 2024-10-17
Update LIT1293Dersan.xml

MarcinKrawczuk:


eu-genia:
It is OK to close without merging since the file had been deleted and replaced by LIT2487Vision in the meanwhile, but it is a pity if the additions/changes go lost, so I compared the files and transferred the changes to LIT2487Vision. Best to not simply delete but check if anything goes lost, thank you!

PR #1260, opened 2024-10-18
2024 10 11 jk qene alam

karljonaskarlsson:
The last qene in the collection in Uppsala, if I have not missed any! Images available here: https://www.alvin-portal.org/alvin/imageViewer.jsf?dsId=ATTACHMENT-0294&pid=alvin-record:280997

### DIFF
--- a/new/LIT7098QeneAlam.xml
+++ b/new/LIT7098QeneAlam.xml
@@ -73,7 +73,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     The first line was originally laid out as four lines in the manuscript, each ending with ፤ .
                 However, later this was noted and the lines were marked as constituting a single line by a speaker of
                 Swedish (who noted down <foreign xml:lang="sv">"1 rad"</foreign>, "one line").
-                Judging from the rhyme, lines 4 and 5 also appear to have been divided—each into two lines—but this has not
+                Judging from the rhyme, lines 3, 4, and 5 also appear to have been divided—each into two lines—but this has not
                 been noted in the manuscript.</note>
                 <ab>
                     <l n="1">ዓለም፡ እስከ፡ ጊዜሃ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice>
@@ -81,7 +81,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         ወታወርደኒ፡ ካእበ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
                                 <del rend="overUnderlined">ሐ</del><add place="above">ኀ</add>በ፡ ዘቀዳሚ፡ መርገም፤</l>
                     <l n="2">ለዘተወለድነ፡ በሥጋ፡ ወደም፤</l><pb n="2r"/>
-                    <l n="3">ለሕልፈት፡ ዛቲ፡ ዓለም፤</l>
+                    <l n="3">ዘአእመራሰ፡ ጠቢብ፡ <sic>ላባዊ</sic><choice><sic>፤</sic><corr resp="JK">፡</corr></choice>
+                        ለሕልፈት፡ ዛቲ፡ ዓለም፤</l>
                     <l n="4">ፍ<choice><sic>ደ</sic><corr resp="JK">ዳ</corr></choice>፡ 
                         ወይን፡ ይስተይ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
                         ማየ፡ ዝናም፤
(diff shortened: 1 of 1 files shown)


new/LIT7098QeneAlam.xml
```diff
@@ -0,0 +1,97 @@
...
+                However, later this was noted and the lines were marked as constituting a single line by a speaker of
+                Swedish (who noted down <foreign xml:lang="sv">"1 rad"</foreign>, "one line").
+                Judging from the rhyme, lines 4 and 5 also appear to have been divided—each into two lines—but this has not
+                been noted in the manuscript.</note>
+                <ab>
+                    <l n="1">ዓለም፡ እስከ፡ ጊዜሃ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice>
+                        በሐብለ፡ ዝንጋኤ፡ ትስሕበኒ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice>
+                        ወታወርደኒ፡ ካእበ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
+                                <del rend="overUnderlined">ሐ</del><add place="above">ኀ</add>በ፡ ዘቀዳሚ፡ መርገም፤</l>
+                    <l n="2">ለዘተወለድነ፡ በሥጋ፡ ወደም፤</l><pb n="2r"/>
```
CarstenHoffmannMarburg, line 83: One line in the manuscript is missing here: ዘከእመራሰ፡ ጠቢብ፡ ላባዊ፤

new/LIT7098QeneAlam.xml
```diff
@@ -0,0 +1,97 @@
...
+                    <l n="3">ለሕልፈት፡ ዛቲ፡ ዓለም፤</l>
+                    <l n="4">ፍ<choice><sic>ደ</sic><corr resp="JK">ዳ</corr></choice>፡ 
+                        ወይን፡ ይስተይ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
+                        ማየ፡ ዝናም፤
+                    </l>
+                    <l n="5">ወህየንተ፡ ልብስ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
+                        ይትከደን፡ ቈጽለ፡ አእዋም፤
+                    </l>
+                    <l n="6">እስመ፡ ዛቲ፡ ጽላሎት፡ ወሕልም(ት)።</l>
```
CarstenHoffmannMarburg, line 92: I think, the brackets are clearly by a later hand. They should be seen as a way to delete the final ት to bring the verse in accordance with the rhyme.

CarstenHoffmannMarburg (changes requested):

new/LIT7098QeneAlam.xml
karljonaskarlsson: You are right! Thanks!

new/LIT7098QeneAlam.xml
karljonaskarlsson: Yes, but how to encode it? Check if you think what I did now might work

karljonaskarlsson: @CarstenHoffmannMarburg , if you think this works, maybe this file can be approved?

new/LIT7098QeneAlam.xml
```diff
@@ -73,23 +73,24 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                        ለሕልፈት፡ ዛቲ፡ ዓለም፤</l>
                     <l n="4">ፍ<choice><sic>ደ</sic><corr resp="JK">ዳ</corr></choice>፡ 
                         ወይን፡ ይስተይ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
                         ማየ፡ ዝናም፤
                     </l>
                     <l n="5">ወህየንተ፡ ልብስ<choice><sic>፤</sic><corr resp="JK">፡</corr></choice> 
                         ይትከደን፡ ቈጽለ፡ አእዋም፤
                     </l>
-                    <l n="6">እስመ፡ ዛቲ፡ ጽላሎት፡ ወሕልም(ት)።</l>
+                    <l n="6">እስመ፡ ዛቲ፡ ጽላሎት፡ ወሕልም<add place="inline">(</add>ት<add place="inline">)</add>።</l>
```
CarstenHoffmannMarburg, line 93: I think, it is not added something to the text, but deleted by a later hand:
`ወሕልም<del rend="encircled">ት</del>።`

If you want to make it more explicit, that the ት is put in brackets, but not in a true circle, you need to express it in a note.

karljonaskarlsson: It's put in parentheses, which is not necessarily the same as deleting, as I see it. It's clearly a "Latin" scribal practice and sign used in an Ethiopian-script context. Maybe the person who added the parentheses (this may have been someone in Sweden) just wanted to express that the ት disturbs the rhyme, but it's still there? I would prefer to mark this unusual scribal practice not as `<del>`

CarstenHoffmannMarburg (approved):

PR #1263, opened 2024-10-22
Update LIT7052 according to Jonas requests

CarstenHoffmannMarburg:
Second attempt to merge these two littly changes.


eu-genia (approved):
As far as I can see this is the first pull request from this branch.
All good.

PR #1265, opened 2024-10-22
Create LIT7094PrayerDevils.xml

CarstenHoffmannMarburg:
Second attempt to merge this file (merging "stale branches"). Cf. https://github.com/BetaMasaheft/Documentation/issues/2619.

### DIFF
--- a/1001-2000/LIT1171Awdana.xml
+++ b/1001-2000/LIT1171Awdana.xml
@@ -35,7 +35,17 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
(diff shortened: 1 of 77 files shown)


eu-genia:
Again there are NO FILES under the FILES tab. Please check exactly what you are trying to do.

Does this record need to be created?

Does it need to be deleted (if already created and merged)?

Does it exist on betamasaheft.eu but not on github? Then synchronize.

eu-genia:
no it is not temporary it means something is wrong with the file

eu-genia:
or rather another file pointing to this one (there is a duplicate file among those linked to this one as the error says "The actual cardinality for parameter 1 does not match the cardinality declared in the function's signature: string($arg as item()?) xs:string. **Expected cardinality: zero or one, got 2.**")

eu-genia:
LIT7094PrayerDevils got expanded twice, I do not know how this happened yet

PR #1157, opened 2024-07-27
Create LIT7042PrayerMichael

CarstenHoffmannMarburg:
Do not mind the strange name of the file! I was confused for a moment. This record is about Legewon.

PR #1267, opened 2024-10-23
Rename LIT7042

CarstenHoffmannMarburg:
According to Eugenias request in https://github.com/BetaMasaheft/Works/pull/1157#pullrequestreview-2321807087.


eu-genia:
You must change the filename in xml:id

eu-genia:
xml:id must always be the same as the filename

eu-genia (approved):

PR #1290, opened 2024-10-31
Update records concerning prayers against stomach pain

CarstenHoffmannMarburg:
I have seen that LIT4624Prayer and LIT6716PPWegat are essentially the same prayer with the only difference, that LIT6716PPWegat contains ʾasmat. I wanted to indicate this relationship by adding relation in body and by updating the abstracts. 

I do not know, what happened to LIT4705Prayer, which was mentioned in line 65 of LIT4624Prayer. LIT4705Prayer is not in the app. I think, this note can be deleted.

### DIFF
--- a/4001-5000/LIT4624Prayer.xml
+++ b/4001-5000/LIT4624Prayer.xml
@@ -27,6 +27,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                 <listWit>
                     <witness corresp="BAVet77#ms_i7"/>
                     <witness corresp="BAVet98#ms_i7"/>
+                    <witness corresp="BLorient12959#ms_i4"/>
                 </listWit>
             </sourceDesc>
         </fileDesc>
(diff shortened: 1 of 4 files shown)


eu-genia:
As you can easily check through 
https://betamasaheft.eu/newSearch.html
image

the correct ID for 4705 if LIT4705MagicPr

eu-genia:
Please also consider that 4705 has as a comment
```
<ab>
                        ጸሎት፡ በእንተ፡ ሕማመ፡ ውግዓት፡ አላህ፡ በላህ፡ ወላህ፡ ሰተፎን፡ ከተፎን፡
                        <!-- Cp. LIT4624Prayer and LIT4625Prayer with the same title (በእንተ፡ ሕማመ፡ ውግ[ዓ]ት፡) but different incipit-->
                    </ab>
```

CarstenHoffmannMarburg:
> the correct ID for 4705 if LIT4705MagicPr

In this case, I suggest to update LIT4705MagicPr as well as LIT4624Prayer and LIT6716PPWegat, creating relations between them and adding references to their abstracts.

eu-genia (approved):

nafisa-valieva (approved):

PR #1292, opened 2024-11-01
Malkea Natnael (RepCh 112) Malkea Tadewos (RepCh 113)

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/2640

Dear @smaugustine can you please check: 

you have the hymn to Nathaniel as 2827 https://malkeagubae.com/works/2827/ but we have it as 2828 - I have mapped our 2827 to your 2828, maybe you can do the same, at the moment the Bm pointer from your side is wrong 

or rename the record as also the next one has the same problem:

you have the hymn to Thaddeus as 2828 https://malkeagubae.com/works/2828/ but we have it as 2829

(2830 is OK)

if you rename 2827 to 2828 and 2828 to 2829 please let me know so that I can update the `relation`

### DIFF
--- a/2001-3000/LIT2828RepCh112.xml
+++ b/2001-3000/LIT2828RepCh112.xml
@@ -4,7 +4,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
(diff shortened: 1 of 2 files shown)


smaugustine:
Sorry @eu-genia, you're right, it was a oversight on my part trying to correlate all the numbers between Chaine and the Clavis. I'll push a correction now that will replace the numbers.

eu-genia:
Sorry, 2828 has become 2829, but 2827 is still there, should be 2828
image

PR #1299, opened 2024-11-05
2024 10 25 jk prayer childbirth

karljonaskarlsson:
Created file according to https://github.com/BetaMasaheft/Documentation/issues/2660

In the issue preceding the creation of this file, @CarstenHoffmannMarburg suggested to record relations to two prayers in other manuscripts that, as far as I understand, are not very well described. I have suggested a way of encoding this but would be grateful if someone with more experience says if this works or not. Thank you!

### DIFF
--- a/new/LIT7121PrayerChildbirth.xml
+++ b/new/LIT7121PrayerChildbirth.xml
@@ -59,8 +59,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="LIT7121PrayerChildbirth" passive="UppEt29"/>
                     <relation name="ecrm:P129_is_about"  active="LIT7121PrayerChildbirth" passive="PRS8083Raphael"/>
-                    <relation name="dcterms:relation " active="LIT7121PrayerChildbirth" passive="EMIP03239#147r"/>
-                    <relation name="dcterms:relation " active="LIT7121PrayerChildbirth" passive="EMIP03223#202v"/>
+                    <relation name="dcterms:relation" active="LIT7121PrayerChildbirth" passive="EMIP03239#147r"/>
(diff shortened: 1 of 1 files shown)


nafisa-valieva (approved):

MassimoVilla (approved):

new/LIT7121PrayerChildbirth.xml
```diff
@@ -0,0 +1,80 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7121PrayerChildbirth" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ጸሎት፡ በእንተ፡ ሕማመ፡ ወሊድ፡</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Ṣalot baʾənta ḥəmāma walid</title>
+                <title xml:lang="en" corresp="#t1">Prayer for the pain of birthgiving</title>
```
CarstenHoffmannMarburg, line 10: We used to translate baʾənta with "against", if it something negative, even if the preposition "for" is idiomatically perfectly OK in this context.

CarstenHoffmannMarburg (approved):

new/LIT7121PrayerChildbirth.xml
karljonaskarlsson: Maybe "against" is even more idiomatic here, though! Thanks, I will change it :)

eu-genia (approved):

PR #1305, opened 2024-11-08
Lit7134 story

MarcinKrawczuk:
I'm not giving an Ethiopic title, if you look at Zarzeczny's edition there's nothing which would resemble a title.

### DIFF
--- a/new/LIT7134Story.xml
+++ b/new/LIT7134Story.xml
@@ -42,8 +42,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             </textClass>
             <langUsage>
                 <language ident="en">English</language>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
> I'm not giving an Ethiopic title, if you look at Zarzeczny's edition there's nothing which would resemble a title.

I think, it is alright to have an English title in this case.

CarstenHoffmannMarburg:
It would be nice to have some manuscript witnesses, if possible.

CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg:
I think, we have a server down at the moment. Please do not merge, before checking that the bm website is online!

PR #1315, opened 2024-11-12
Lit1803 lifeof corr mk 12.11.24

MarcinKrawczuk:
The list of witnesses does not display in the app, maybe this little change will help.

### DIFF
--- a/1001-2000/LIT1803Lifeof.xml
+++ b/1001-2000/LIT1803Lifeof.xml
@@ -58,6 +58,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change who="ES" when="2016-02-09">CREATED: text record</change>
          <change who="DR" when="2017-03-14">Added keywords and relation</change>
          <change who="MV" when="2017-11-20">Updated</change>
+         <change who="MKr" when="2024-11-12">Corrected listWit</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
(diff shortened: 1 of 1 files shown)


1001-2000/LIT1803Lifeof.xml
```diff
@@ -31,7 +31,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </sourceDesc>
          <sourceDesc>
             <listWit>
-               <witness corresp="BNFabb60"/>
+               <witness corresp="BNFabb60">BnF d'Abb. 60, f. 1r-16r</witness>
```
MassimoVilla, line 34: Thank you Marcin. This is very good in principle, just I do not know if this way of encoding is used elsewhere. Perhaps @eu-genia can tell something more.
In any case, what about `<witness corresp="BNFabb60#i1"/>`? Does this point to the exact textual unit?

eu-genia: `<witness corresp="BNFabb60#i1"/>` would not print anything ) first, `i1 `does not exist, second, the app can only print the exact shelfmark if only the exact shelfmark is in `corresp`, if you want to link to an item you must provide the text 

so it is OK to list
`<witness corresp="ID"/> `
or 
`<witness corresp="ID#ms_i1">ms Full shelfmark, ff. xx-yy</witness>`

In this case the manuscript does not appear only because there is no item BNFabb60 in the application. the ID must be corrected to point to an existing record

eu-genia: (or a record for BNFabb60 must be created, then no correction in this LIT record would be necessary at all)

MassimoVilla (approved):

eu-genia (approved):
this will make the text visible but the ms as far as I can see is still not in the system, there is neither BNFabb60 nor any msItem inside, so the link would point to nowhere for the time being

MassimoVilla:
I'm creating the MS record right now, with minimal info on the textual content

PR #1322, opened 2024-11-15
Create records for the Maṣḥafa falāsfā ṭabibān

MassimoVilla:
Following https://github.com/BetaMasaheft/Documentation/issues/2686#event-15281654086, I have created two new records for the short and long recensions of the Maṣḥafa falāsfā ṭabibān.

There's still much to do with single textparts, but this probably requires a specific and thorough effort because the order and presence of individual sentences seem to vary and must be checked against the single manuscripts, since there's no edition to which to refer.

MS records have been updated here: https://github.com/BetaMasaheft/Manuscripts/pull/2707

### DIFF
--- a/1001-2000/LIT1925Mashaf.xml
+++ b/1001-2000/LIT1925Mashaf.xml
@@ -58,7 +58,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                Collection of sentences and sayings of Greek philosophers, Fathers of the Church and biblical figures. Some apophthegms are anonymous.
                The work was translated from the Arabic <foreign xml:lang="ar">Kitāb al-bustān wa-qāʿidat al-ḥukamāʾ wa-šams al-ʾādāb</foreign> between
                1510 and 1522 by <persName role="translator" ref="PRS7077Mikael"/> and enjoyed much popularity given the high number of manuscripts.
-               There exist a short and a long recension of the work. The <ref type="work" corresp="LIT7144MFTabiban">short recension</ref> generally
+               There is a short and a long recension of the work. The <ref type="work" corresp="LIT7144MFTabiban">short recension</ref> generally
                preserves the Arabic text. The <ref type="work" corresp="LIT7145MFTabiban">long recension</ref> includes additional sayings of the Church Fathers
                and other stories, especially of <persName ref="PRS1499Ahiqar"/>.</p>
             <p>
--- a/1001-2000/LIT1925Mashaf.xml
+++ b/1001-2000/LIT1925Mashaf.xml
@@ -48,11 +48,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          </xi:include>
       </encodingDesc>
       <profileDesc>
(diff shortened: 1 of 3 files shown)


eu-genia (approved):
Thank you, very helpful!

1001-2000/LIT1925Mashaf.xml
```diff
@@ -48,11 +48,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+         </langUsage>
          <abstract>
-            <p>There exist a short and a long recension of the <foreign xml:lang="gez">Maṣḥafa falāsfā ṭabibān</foreign>. The short recension generally
-            preserves the Arabic text. To the long recension, sayings of the Church fathers and other stories,
-            especially of <persName ref="PRS1499Ahiqar"/>, have been added.</p>
+            <p>
+               Collection of sentences and sayings of Greek philosophers, Fathers of the Church and biblical figures. Some apophthegms are anonymous.
+               The work was translated from the Arabic <foreign xml:lang="ar">Kitāb al-bustān wa-qāʿidat al-ḥukamāʾ wa-šams al-ʾādāb</foreign> between
+               1510 and 1522 by <persName role="translator" ref="PRS7077Mikael"/> and enjoyed much popularity given the high number of manuscripts.
+               There exist a short and a long recension of the work. The <ref type="work" corresp="LIT7144MFTabiban">short recension</ref> generally
```
CarstenHoffmannMarburg, line 61: There **is** a short and a long recension of the work.

new/LIT7144MFTabiban.xml
```diff
@@ -0,0 +1,121 @@
...
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>
+                    This is the short recension of the <foreign xml:lang="gez">Maṣḥafa falāsfā ṭabibān</foreign>.
```
CarstenHoffmannMarburg, line 36: I would suggest to add a reference to LIT1925Mashaf to the abstract as well, because in the relation, it is hardly visible to the users.

CarstenHoffmannMarburg (approved):

1001-2000/LIT1925Mashaf.xml
nafisa-valieva, line 1: Massimo, if it is not too much to ask, from the experience, we belive that explcits are also very needed!

nafisa-valieva (approved):
Massimo, thank you! if it is not too much to ask, from the experience, we belive that explcits are also very needed!

eu-genia:
> Massimo, thank you! if it is not too much to ask, from the experience, we belive that explcits are also very needed!
We are grateful for any data, incipits are already great.

Nafisa, we do what we can. Our primary goal is to encode what is given in the catalogue. If the catalogue provides incipit and explicit - yes. If the catalogue only gives incipit - we go with the incipit. We are not re-cataloguing unless this is specifically requested (not the case for Grebaut Tisserant). For works - whatever anyone feels like providing is great as long as it helps disambiguation.

Let us try to be as economical as possible, thank you!

Massimo, thank you, you do not need to provide explicits.

PR #1325, opened 2024-11-25
Updated transcription in titles

karljonaskarlsson:
Updated transcription in titles


eu-genia:
Nb the files are many so I will have to update the DB manually, please do not merge yet.
I will also update the list then.
Thank xou!

PR #1249, opened 2024-10-16
Update LIT5673AkkonuColl.xml

GueshSol:
Added elements on the Miracle of Mary and Isaac

PR #1329, opened 2024-11-28
Modifications on the generic miracle record 3592 ( Miracle of Mary: St Mary and Isaac)

GueshSol:
Following my discussion with , I believe @eu-genia it is essential to create two CAe IDs for the generic miracle record, LIT3592Miracle. This record is linked to several manuscripts, making it difficult to modify. Therefore, I propose to leave the generic record (only modified the realtions and added note in the abstract on the differences) and create two CAe IDs that correspond to it, each providing the specific manuscript attestations for the two related but distinct miracles. I will provide the newly created CAe IDs soon. In the meantime, I would appreciate your thoughts on modifying this generic work record.

### DIFF
--- a/3001-4000/LIT3592Miracle.xml
+++ b/3001-4000/LIT3592Miracle.xml
@@ -6,9 +6,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
       <fileDesc>
          <titleStmt>
             <title xml:id="t1" xml:lang="en">Miracle of Mary: Miracle of the monk Isaac</title>
-            <title xml:id="t2" xml:lang="en">Miracle of the monk Yǝsḥāq</title>
+            <title xml:id="t2" xml:lang="en">Miracle of Mary: Miracle of the priest Isaac</title>
             <title xml:id="t3" xml:lang="en">The monk, Yǝsḥaq, who prayed for seven years that St
                Mary would appear to him</title>
+            <title xml:id="t4" xml:lang="en">The priest and Abbot, Yǝsḥaq, who celebrated Marian
+               feast anually, but later migrated to another monastery due to financial problems</title>
          </titleStmt>
          <publicationStmt>
             <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
If I understand you correctly, you want to keep one generic record (LIT3592Miracle) and you want to create two new LIT ID for those attestations, that you can describe more precisely. This is a good idea.

3001-4000/LIT3592Miracle.xml
```diff
@@ -25,17 +29,30 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+            href="https://raw.githubusercontent.com/BetaMasaheft/Documentation/master/prefixDef.xml">
             <xi:fallback>
                <p>Definitions of prefixes used.</p>
             </xi:fallback>
          </xi:include>
       </encodingDesc>
       <profileDesc>
          <abstract>
-            <p/>
+            <p> A miracle conflation appears in various manuscripts, linked to this generic miracle
```
CarstenHoffmannMarburg, line 41: I suggest a modification of this abstract. You should 1) give a brief summary of the content and 2) inform the reader about the two existing versions and refer to them, 3) make clear, that this CAe is a general record and meant only for those attestations for which the exact version is not clear.

3001-4000/LIT3592Miracle.xml
```diff
@@ -92,81 +112,25 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                     ብእሲ፡ መነኮስ፡ ቅዱስ፡ ወእምንእሱ፡ ነበረ፡ ህየ፡ ወስሙ፡ ይስሐቅ።
-                 </ab>
-             </div>
             <div type="textpart" subtype="incipit" xml:lang="gez">
-               <note>This incipit is taken from <ref type="mss" corresp="EMML4205"/>.</note>
-               <ab>
-                  ወሀሎ፡ አሐዱ፡ መነኮስ፡ ኄር፡ ዘስሙ፡ ይስሐቅ፡ ወያፈቅራ፡ ለእግዝእትነ፡ ማርያም፡
-                  ወእምንእሱ፡ ነበረ፡ በአሐቲ፡ ደብር፡ በመዋዕሊሁ፡ ለአባ፡ ገብርኤል።
-               </ab>
+               <note>This incipit, related to the first miracle version is taken from <ref
```
CarstenHoffmannMarburg, line 122: I think, these incipits should go the newly created CAe to which they belong. If this is a generic record, it should not contain any incipit or explicit or any other text. Otherwise the encoders and users will be confused about the correct CAe.

CarstenHoffmannMarburg (changes requested):

eu-genia:
It would be much nicer to keep all records together in the same branch https://github.com/BetaMasaheft/Works/pull/1249 as the changes are interconnected and all refer to the same issue. It is very difficult to jump between branches and pull requests for one thing

eu-genia:
Can you move the changes proposed here to the branch https://github.com/BetaMasaheft/Works/pull/1249 and have all in one pull request? then I would close this one

PR #1332, opened 2024-11-30
added textparts to LIT7107Divination

MassimoVilla:
Added textparts to LIT7107Divination, following https://github.com/BetaMasaheft/Documentation/issues/2704

### DIFF
--- a/new/LIT7107Divination.xml
+++ b/new/LIT7107Divination.xml
@@ -45,6 +45,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         </profileDesc>
         <revisionDesc>
             <change who="MV" when="2024-10-21">Created entity</change>
+            <change who="MV" when="2024-11-30">Added textparts</change>
         </revisionDesc>
     </teiHeader>
     <text>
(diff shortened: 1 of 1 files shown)


new/LIT7107Divination.xml
```diff
@@ -150,11 +186,36 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                        <ab>
+                            ስመ፡ ወእመ፡ በበ<supplied reason="omitted">፡</supplied> ፲ ግድፍ፡ ፩ ንጉሥ፡ አልጱስስ፡ ወሀብቱ፡ 
+                            <sic>ሀብቱ፡</sic> ሀለወት፡ ተፈልጠ፡ እምላዕሌሁ፡ መንገለ፡
+                            ምሥራቅ፡ ኢትመጽእ፡ ፍጡነ፡
+                        </ab>
+                    </div>
+                </div>
+                
+                <div type="textpart" xml:id="fenot">
+                    <label>Divinatory chapter entitled <foreign xml:lang="gez">ሐሳበ፡ ፍኖት፡</foreign>, or "Calculation of the paths"</label>
```
CarstenHoffmannMarburg, line 209: Personally, I tend to give shorter labels and avoid any markup in labels. However, I see, that this has been okay for the other textparts already existing before, so I guess, that it is acceptable also for the new ones.

CarstenHoffmannMarburg (approved):

new/LIT7107Divination.xml
MassimoVilla: I can shorten the title deleting "Divinatory chapter entitled", here and in the other textparts. What do others think?

CarstenHoffmannMarburg: I propose to shorten the label and put the rest of the information in a note. However, I think, it would be useful to do that in all the textparts to keep it consistent, if this is not too cumbersome.

eu-genia (approved):

PR #1375, opened 2025-01-07
Updating Magic Prayers on colic

CarstenHoffmannMarburg:
I updated the records for Magic prayers on colic according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2746.

I made a minor correction in LIT5690MagicPrayer

### DIFF
--- a/4001-5000/LIT4595Salot.xml
+++ b/4001-5000/LIT4595Salot.xml
@@ -5,6 +5,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">ጸሎት፡ በእንተ፡ ሕማመ፡ ቍርጸት፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Ṣalot baʾǝnta ḥǝmāma qʷǝrṣat</title>
+                <title xml:lang="en" corresp="#t1">Prayer against the pain of colic</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="DR"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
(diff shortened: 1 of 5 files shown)


eu-genia:
I understood it from the issue that you wanted to delete several, and that 22 and 23 were doublets?

CarstenHoffmannMarburg:
> I understood it from the issue that you wanted to delete several, and that 22 and 23 were doublets?

Yes, my original plan was to delete three records and keep one as generic ID. However, Jonas discovered, that LIT4622Prayer and LIT4623Prayer have longer and different incipits. I discovered later, that also LIT4595Salot and LIT4604Salot might be different, because the former was described as a "mixture of Amharic and Geez". So I changed my plan and updated the records accordingly.

nafisa-valieva (commented):
Thank you! there is no bibliography at all or I missed it?

karljonaskarlsson (approved):

CarstenHoffmannMarburg:
> Thank you! there is no bibliography at all or I missed it?

Correct. I added the relations in a div with the specification "bibliography" as I usually do. If this is incorrect, I can delete the type "bibliography"

```
<div type="bibliography">
                <listRelation>
                    <relation name="dc:relation" active="LIT4622Prayer" passive="LIT4595Salot"><desc>Thematically similar to 
                        <ref type="work" corresp="LIT4595Salot"/>.</desc></relation>
                    <relation name="dc:relation" active="LIT4622Prayer" passive="LIT4604Salot"><desc>Thematically similar to 
                        <ref type="work" corresp="LIT4604Salot"/>.</desc></relation>
                </listRelation>
            </div>
```

CarstenHoffmannMarburg:
> As you know, relations are not visible from the main view, only when one clicks on relations tab which most users do not do for a first impression, so everything that helps understanding and disambiguation is better to be said where it can be seen at once. This is a general comment valid for all work, person, place records.

I understand. I think, it is helpful for further disambiguation in the future. In this case, the basic content is visible from the title and from the keywords. Creating an abstract would not be an advantage, because I do not have any additional information.

eu-genia (approved):

PR #1380, opened 2025-01-09
Qene abb145 emena walatta

DenisNosnitsin1970:


CarstenHoffmannMarburg:
It should be updated to 'mi bazḫu-type' (cf. remark by Jonas in https://github.com/BetaMasaheft/Works/pull/1389).

DenisNosnitsin1970:
ok. thanks!

PR #1388, opened 2025-01-12
Update LIT5690MagicPrayer.xml

CarstenHoffmannMarburg:
I updated this record, because I was not content with t1 and wanted to create an abstract instead. If I remember it correctly, we should not provide Russian, French or Italian translations of Geez titles.

### DIFF
--- a/5001-6000/LIT5690MagicPrayer.xml
+++ b/5001-6000/LIT5690MagicPrayer.xml
@@ -4,15 +4,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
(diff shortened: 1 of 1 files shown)


thea-m (changes requested):
Thank you, very helpful! Could you keep the word "Tiberias" in some searchable way (in the title or abstract) in the record? At least I search often by it to look for this prayer.

nafisa-valieva (approved):
Daria´s work is gone :-( I wonder whether it is a good strategy, since it was correct, but I am ok with the changes.

CarstenHoffmannMarburg:
> Thank you, very helpful! Could you keep the word "Tiberias" in some searchable way (in the title or abstract) in the record? At least I search often by it to look for this prayer.

I updated the abstract and marked up the word in the edition of the incipit.

thea-m (approved):

PR #1389, opened 2025-01-12
Create LIT7237QeneFequrana.xml

CarstenHoffmannMarburg:
Several features are problematic in this Qene in BNFet145. 

- [ ] The first problem is the verse break after መፍርህ, which disturbs the rhyme that is always in ል in this poem. For this reason, I have deleted this verse break and moved it to ሣህል, which is two words before. 
- [ ] The second problem is the addition of ጊዜ፡ ሞቶሙ፡ ረድኤት፡  and ወ, which was written between the lines. I am not quite sure, if I have put these words in the right place in verse 1. You can check, if this verse sounds smooth to you. 
- [ ] The third problem is the translation of the title. I have translated it as a circumstantial clause similar to an Arabic ḫāl-sentence. Perhaps you have another suggestion. It may be better to include more words from the first verse in the title. 
- [ ] The fourth is the qene type. If I understand Guidi correctly, all poems with eight verses are mawaddes while nine verses make up a mawaddes kwellekemu. It does not sound very logical to me, but my Italian is too weak to be sure about Guidi's statement. Please check and confirm my assumption (also relevant for LIT7223QeneMenase and other records of mawaddes kwellekemu). 
- [ ] Fifth I did not follow Nafisa's reading of ኢሎፍሲ፡ and preferred ኢሎፍሊ፡, because I think it is about 'Philistines'.

### DIFF
--- a/new/LIT7237QeneFequrana.xml
+++ b/new/LIT7237QeneFequrana.xml
@@ -1,0 +1,77 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7237QeneFequrana" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ፍቁራነ፡ ኮኑ፡ ወኢተሴሰዩ፡</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Fǝqurāna konu wa-ʾitasessayu (Qǝne of the mawaddǝs-kʷǝllǝkǝmu-type)</title>
+                <title xml:lang="en" corresp="#t1">They were happy even though they were not fed…</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <listWit>
+                    <witness corresp="BNFabb145"/>
+                </listWit>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Neither type nor author of the qǝne are indicated in the manuscript. According to the classification by 
+                    <bibl><ptr target="bm:Guidi1900Qene"/><citedRange unit="page">4-5</citedRange></bibl> a qǝne of the 
+                    mawaddǝs-kʷǝllǝkǝmu-type.</p>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
+                    <term key="Poetry"/>
+                    <term key="Qene"/>
+                    <term key="Kwellekomu"/>
+                </keywords>
+            </textClass>
+            <langUsage>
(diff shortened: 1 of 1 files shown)


new/LIT7237QeneFequrana.xml
```diff
@@ -0,0 +1,77 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7237QeneFequrana" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ፍቁራነ፡ ኮኑ፡ ወኢተሴሰዩ፡</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Fǝqurāna konu wa-ʾitasessayu (Qǝne of the mawaddǝs-kʷǝllǝkǝmu-type)</title>
```
DenisNosnitsin1970, line 9: Yes, it seems to be meweddes. Meweddes is a complex type and Guidi presents it also in a complex way.  I also understand that simpe meweddes is eight verses and meweddes kwellekemu is nine verses.

DenisNosnitsin1970 (changes requested):

new/LIT7237QeneFequrana.xml
```diff
@@ -0,0 +1,77 @@
...
+            <change who="CH" when="2025-01-12">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="edition" xml:lang="gez">
+                <note>This edition is based on <ref type="mss" corresp="BNFabb145"/>, fol. 56ra-b. Division into verses is indicated with ። . 
+                    The end of the poem is indicated with ። ። .</note>
+                <ab>
+                    <l n="1">ፍቁራነ፡ ኮኑ፡ ወኢተሴሰዩ፡ <add place="above">ጊዜ፡ ሞቶሙ፡ ረድኤት፡ ወ</add>ጊዜ፡ ሕይወቶሙ፡ ሣህል<supplied reason="omitted">።</supplied></l>
```
DenisNosnitsin1970, line 64: The place of the addition appears ok for me.

DenisNosnitsin1970 (changes requested):

new/LIT7237QeneFequrana.xml
karljonaskarlsson, line 9: If we want to be consistent in English, I think "mawaddǝs kʷǝllǝkǝmu-type" is preferrable to "mawaddǝs-kʷǝllǝkǝmu-type". Also below.

karljonaskarlsson, line 9: I would transliterate "wa-ʾi-tasesayu"

new/LIT7237QeneFequrana.xml
```diff
@@ -0,0 +1,77 @@
...
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Neither type nor author of the qǝne are indicated in the manuscript. According to the classification by 
+                    <bibl><ptr target="bm:Guidi1900Qene"/><citedRange unit="page">4-5</citedRange></bibl> a qǝne of the
```
karljonaskarlsson, line 38: Maybe "…, it is a qǝne of the…"

new/LIT7237QeneFequrana.xml
karljonaskarlsson: Me too! :)

karljonaskarlsson (approved):
Very nice! :)

karljonaskarlsson:
I agree with you about reading ኢሎፍሊ

eu-genia (approved):

new/LIT7237QeneFequrana.xml
CarstenHoffmannMarburg: This is also my idea of English syntax. However, in this case, we should change 'mi-bazḫu-type' to 'mi bazḫu-type', as well (cf. LIT7225EklaAbasaha in https://github.com/BetaMasaheft/Works/pull/1380/files).

new/LIT7237QeneFequrana.xml
CarstenHoffmannMarburg: Do we want to be consistent in all our qəne records - then I should update the other records, too.

new/LIT7237QeneFequrana.xml
eu-genia: no, in the case of mi-bazḫu the dash has a different meaning (not linking two words but separating morphological units)

new/LIT7237QeneFequrana.xml
karljonaskarlsson: I have used a different phrase for this, I believe. But you can keep this formulation – in case someone want to change it in the future, it will still be a manageable task :)

CarstenHoffmannMarburg:
@DenisNosnitsin1970 - If all of your requests have been solved, can you approve this pull request?

DenisNosnitsin1970 (approved):

PR #1407, opened 2025-01-23
Update LIT1340EnochE.xml

eu-genia:
The IDs `#FSUrueppII1` and `#SHOr271a` are not valid witness IDs. The correct IDs are `#fsurueppII1` and `#shor271a` (capitalization is significant) and both are already present in the list.


eu-genia:
As mentioned, all inconsistencies in display are due to issues with the XML records.

nafisa-valieva (approved):
Thank you very much!

PR #1423, opened 2025-01-29
Create LIT7260PPKumael.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2787.

### DIFF
--- a/new/LIT7260PPKumael.xml
+++ b/new/LIT7260PPKumael.xml
@@ -7,7 +7,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <titleStmt>
                 <title xml:lang="en" xml:id="t1">Prayer for terrifying devils</title>
                 <title xml:lang="gez" xml:id="t2">ኩማኤል፡ ፫ጊዜ፡ ኽኤል፡</title>
-                <title xml:lang="gez" corresp="#t2" type="normalized">Kumāʾel 7-gize Ḵǝʾel</title>
+                <title xml:lang="gez" corresp="#t2" type="normalized">Kumāʾel 3 gize Ḵǝʾel</title>
                 <title xml:lang="en" corresp="#t2">Kumāʾel, 7 times Ḵǝʾel</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="CH"/>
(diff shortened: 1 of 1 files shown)


new/LIT7260PPKumael.xml
```diff
@@ -0,0 +1,67 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7260PPKumael" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Prayer for terrifying devils</title>
+                <title xml:lang="gez" xml:id="t2">ኩማኤል፡ ፫ጊዜ፡ ኽኤል፡</title>
+                <title xml:lang="gez" corresp="#t2" type="normalized">Kumāʾel 7-gize Ḵǝʾel</title>
```
karljonaskarlsson, line 10: Is the hyphen needed?

karljonaskarlsson (approved):

new/LIT7260PPKumael.xml
eu-genia: I would always tokenize, as we do with prepositions

eu-genia (approved):

new/LIT7260PPKumael.xml
karljonaskarlsson: Whether hyphen or not, I believe it must be "3" :)

CarstenHoffmannMarburg: > I would always tokenize, as we do with prepositions (to mark that these are two separate morphemes even if orthographically attached)

I understand your remark, that you recommend to write it without hyphen: Kumāʾel 3 gize Ḵǝʾel

karljonaskarlsson (approved):

new/LIT7260PPKumael.xml
eu-genia: I actually meant we use hyphen to mark separate morphemes but one orthographical unit, also with numerals, but ok

PR #1435, opened 2025-02-06
Nv 04.01.2025 wp

nafisa-valieva:
As suggested by Eugenia [here](https://github.com/BetaMasaheft/Manuscripts/pull/2622) , I have created two missing records.

### DIFF
--- a/1001-2000/LIT1509GadlaW.xml
+++ b/1001-2000/LIT1509GadlaW.xml
@@ -60,6 +60,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change who="ES" when="2016-02-09">CREATED: text record</change>
          <change who="MV" when="2017-11-07">Edited</change>
          <change who="MV" when="2019-03-07">added content</change>
+         <change when="2025-02-06" who="NV">modified file as cointing only gadl</change> 
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
@@ -92,7 +93,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <div type="edition">
                <div type="textpart" subtype="incipit" xml:id="Introduction">
                   <label>Introduction</label>
-                    
+                  
                </div>
                
                <div type="textpart" subtype="Gadl" xml:id="Gadl">
@@ -102,21 +103,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   </div>
                   
                </div>
-               <div type="textpart" subtype="Taammer" xml:id="Taammer">
-                 
-                  <label>Taʾammǝr</label>
-                  <div type="textpart" subtype="chapter">
-                     <ab/>
-                  </div>
-               </div>
-               <div type="textpart" subtype="hymn" xml:id="hymn">
-                  <label>hymn</label>
-                  <div type="textpart" subtype="chapter">
-                     <ab/>
-                  </div>
-                  
-               </div>
-               
             </div>
 
       </body>
--- a/1001-2000/LIT1509GadlaW.xml
+++ b/1001-2000/LIT1509GadlaW.xml
@@ -45,7 +45,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <profileDesc>
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
          <abstract><p>
-            Although no manuscript containing only the vita (gadl) has been reported thus far, this record pertains solely to the vita. It is distinguished from the miracles, as the set of miracles following the vita can vary. Furthermore, the gadl and miracles are visually and logically distinct.
+             The Life of <persName ref="PRS9848Walatta"/>, usually followed by a set of <ref type="work" corresp="LIT7073TammeraW">Miracles</ref>. 
+            Although no manuscript containing only the vita (gadl) has been reported thus far, it is distinguished from the <ref type="work" corresp="LIT7073TammeraW">miracles</ref>, as the set of miracles following the vita can vary. Furthermore, the gadl and miracles are visually and logically distinct.
          </p></abstract>
          <textClass>
             <keywords>
--- a/1001-2000/LIT1509GadlaW.xml
+++ b/1001-2000/LIT1509GadlaW.xml
@@ -89,6 +89,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   <ptr target="bm:Papi1943WalattaPetros"/>
                </bibl>
             </listBibl>
+            <listRelation>
+               <relation name="dcterms:isPartOf" active="LIT1509GadlaW" passive="LIT7229DossierWP"></relation>
+               <relation name="ecrm:P129_is_about" active="LIT1509GadlaW" passive="PRS9848Walatta"/>
+            </listRelation>
          </div>
             <div type="edition">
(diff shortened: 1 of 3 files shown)


eu-genia:
Copying here the discussion from the manuscript PR as otherwise impossible to understand

> here and in BLorient730 - we need a separate CAe ID for the hagiographic dossier of Walatta Petros which would have relations to its parts (cp. e.g. LIT1008DossierTH, LIT6172DossierGeorge, etc)

> So that we always have, for hagiographical cycles separate IDs for 

```
DOSSIER
-gadl
-miracles
-hymns
-relics, commemorations, other texts that may form part of the collection

```

> so in this case

```
**new ID for the set which contains
--LIT1509GadlaW = if Vita and Miracles are always together and without variation in the number and order of miracles we can leave is as it is with Vita and Miracles, if there may be any variation or if you are not sure then Miracles must have an extra ID, too
--LIT3001RepCh281 = Malk
--LIT2813RepCh97
--as well as possibly LIT3072RepCh352, LIT2892RepCh173, LIT2720RepCh4

```

> here and elsewhere then the nesting should be

```
ms_i1 = dossier
  ms_i1.1 = gadl
  ms_i1.2 = miracles
  ms_i1.3 = malk1
  ms_i1.4 = malk2
  ....
ms_i2, 3, 4 = other possible texts 
```

_Originally posted by @eu-genia in https://github.com/BetaMasaheft/Manuscripts/pull/2622#discussion_r1768394861_

new/LIT7229DossierWP.xml
```diff
@@ -0,0 +1,76 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7229DossierWP" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Hagiographic Dossier of St Walatta Ṗeṭros</title>
```
eu-genia, line 8: en

eu-genia (commented):
Very nice!
Probably LIT1509GadlaW  must be edited to clarify that it should be used only for the Vita, not for the Vita+Miracles combination; pointers should be checked

It is Ok for the moment but if the miracles set is not stable and may vary from ms to ms, each miracle then has an individual circulation and thus needs its own CAe ID, as it has been done for other miracles, e.g. https://github.com/BetaMasaheft/Documentation/issues/2728 https://github.com/BetaMasaheft/Documentation/issues/2719

If we always do have the same set of 11 miracles then it is OK to have them as chapters only

1001-2000/LIT1509GadlaW.xml
```diff
@@ -60,6 +60,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change who="ES" when="2016-02-09">CREATED: text record</change>
          <change who="MV" when="2017-11-07">Edited</change>
          <change who="MV" when="2019-03-07">added content</change>
+         <change when="2025-02-06" who="NV">modified file as cointing only gadl</change>
```
eu-genia, line 63: containing?

1001-2000/LIT1509GadlaW.xml
```diff
@@ -102,21 +103,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   </div>
                   
                </div>
```
eu-genia, line 105: add an abstract that clarifies that this record must be used for the Gadl only

before removing the xml:id you must check that there are no manuscripts pointing to `LIT1509GadlaW#Taammer` or `LIT1509GadlaW#Hymn` and if there are replace the references with the one to the new IDs, or we will have an error

eu-genia (changes requested):

1001-2000/LIT1509GadlaW.xml
eu-genia: I assume BLorient730 must be edited

new/LIT7229DossierWP.xml
```diff
@@ -0,0 +1,76 @@
...
+            <div type="bibliography">
+             
+                <listBibl type="secondary">
+                    <bibl>
+                        <ptr target="bm:Cohen2010WalattaPetros"/>
+                    </bibl>
+                </listBibl>
+                <listRelation>
+                    <relation name="saws:contains" active="LIT7229DossierWP" passive="LIT1509GadlaW">
+                        <desc>The dossier contains the Gadla Walatta Ṗeṭros (Acts)</desc>
```
karljonaskarlsson, line 60: I don't think "the" is needed with these work titles (here and below)

1001-2000/LIT1509GadlaW.xml
```diff
@@ -102,21 +103,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                  <div type="textpart" subtype="chapter">
-                     <ab/>
-                  </div>
-               </div>
-               <div type="textpart" subtype="hymn" xml:id="hymn">
-                  <label>hymn</label>
-                  <div type="textpart" subtype="chapter">
-                     <ab/>
-                  </div>
-
```
karljonaskarlsson, line 117: Are the relations to other entities missing? (Or am I just failing to find them?)

Perhaps:

```
<listRelation>
<relation name="dcterms:isPartOf" active="LIT1509GadlaW" passive="LIT7229DossierWP"></relation>
<relation name="ecrm:P129_is_about" active="LIT1509GadlaW" passive="PRS9848Walatta"/>
</listRelation>
```

karljonaskarlsson (commented):
Nice! :)

1001-2000/LIT1509GadlaW.xml
```diff
@@ -47,6 +44,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       </encodingDesc>
       <profileDesc>
          <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
+         <abstract><p>
+            Although no manuscript containing only the vita (gadl) has been reported thus far, this record pertains solely to the vita. It is distinguished from the miracles, as the set of miracles following the vita can vary. Furthermore, the gadl and miracles are visually and logically distinct.
+         </p></abstract>
```
eu-genia, line 49:
```
The Life of `<persName ref="PRS9848Walatta"/>`, usually followed by a set of <ref type="work" corresp="LIT7073TammeraW">Miracles</ref>. 
Although no manuscript containing only the vita (gadl) has been reported thus far, it is distinguished from the <ref type="work" corresp="LIT7073TammeraW">miracles</ref>, as the set of miracles following the vita can vary. Furthermore, the gadl and miracles are visually and logically distinct.
```

1001-2000/LIT1509GadlaW.xml
nafisa-valieva: done

1001-2000/LIT1509GadlaW.xml
nafisa-valieva: good, I suppose

nafisa-valieva: @karljonaskarlsson, @eu-genia, from my perspective, I think I have addressed your suggestions and remarks. May I ask you to check it once again, and if all is fine, allow the merge?

eu-genia (approved):

PR #1458, opened 2025-02-20
Created three records

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2841#issuecomment-2668610315.

### DIFF
--- a/new/LIT7288PrBaryaDis.xml
+++ b/new/LIT7288PrBaryaDis.xml
@@ -1,0 +1,78 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7288PrBaryaDis" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1"> ጸሎት፡ በእንተ፡ ዓይነተ፡ ባርያ፡ መጋኛ፡ ወሾቶላይ፡ ውግዓት፡ ወቍርጥማት፡ ፍልፀት፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Ṣalot baʾǝnta ʿāynata Bāryā Magāññā wa-Šotolāy wǝgʿāt wa-qʷǝrṭǝmāt
+                    fǝlḍat</title>
+                <title xml:lang="en" corresp="#t1">Prayer against the evil eye of Bāryā, Magāññā, and Šotolāy, chest pain, rheumatism, pleurisy, ...</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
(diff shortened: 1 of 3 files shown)


new/LIT7288PrBaryaDis.xml
```diff
@@ -0,0 +1,78 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Strelcyn reported in <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">136</citedRange></bibl>,
```
karljonaskarlsson, line 36: Maybe this could be syntactically clarified a little? For example:

```
Strelcyn reported in <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">136</citedRange></bibl> 
that this prayer is against blacksmiths, among other things. 
Blacksmiths are usually referred to as <foreign xml:lang="gez">ነሃቢ፡</foreign>, while this
                    text instead uses the term <foreign xml:lang="gez">ጠቢብ፡</foreign>, which can also have many other meanings, such as "wise man / scholar / physician / 
                    artisan / expert / craftsman".
```

new/LIT7289PrChest.xml
```diff
@@ -0,0 +1,80 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7289PrChest" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>    
+                <title xml:lang="gez" xml:id="t1">ጸሎት፡ በእንተ፡ ሕማመ፡ ውግዓት፡ ሊቃኖስ፡ ብሂል፡ ዕፁብ፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Ṣalot baʾǝnta ḥǝmāma wǝgʿāt Liqānos bǝhil ʿǝḍub</title>
+                <title xml:lang="en" corresp="#t1">Prayer against chest pain, which Liqānos called 'the heavy one'</title>
```
karljonaskarlsson, line 10: Normally, I think "bǝhil" means "which is to say", or "i.e.". I've come across it in word lists, where one Arabic word etc. is equated with a Geez word. So an alternative translation (which doesn't make more sense than what you have now) could be: 

Prayer against chest pain. Liqānos means 'harsh'.

But decide what you prefer :)

new/LIT7289PrChest.xml
```diff
@@ -0,0 +1,80 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>The name 'Liqānos' in the title may possibly refer to <persName ref="PRS6318Liqanos"/>.</p>
```
karljonaskarlsson, line 35: Can we really say "title"? Maybe "incipit" is better in this case?

karljonaskarlsson (approved):
Good job! Thank you!

new/LIT7289PrChest.xml
CarstenHoffmannMarburg: I prefer the interpretation, that Liqānos is a person and the last three words are related to the title before. For this reason, I want to stay with the translation above.

new/LIT7289PrChest.xml
CarstenHoffmannMarburg: OK.

PR #1474, opened 2025-02-28
Update LIT7298EncomiumTammeraG.xml

nafisa-valieva:
second merge to make appear the title


eu-genia:
since the list of titles of works in CAe is not any more updated automatically but only manually by myself every now and then https://github.com/BetaMasaheft/Documentation/issues/2766  merges do not influence whether the title is printed or not, it will appear once it is there in the list

I have now updated the list you can check the record

eu-genia (approved):

PR #1439, opened 2025-02-12
Create LIT7277PrDiseases.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2824#issuecomment-2653170792.

PR #1483, opened 2025-03-05
Update according to Magdas requests and remarks

CarstenHoffmannMarburg:
According to your requests and remarks here: https://github.com/BetaMasaheft/Works/pull/1439#pullrequestreview-2660326979


CarstenHoffmannMarburg:
@MagdaKrzyz, can you check and approve my updates?

PR #1502, opened 2025-03-18
Create LIT7273PPEntering.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2803#issuecomment-2624871783.

Apparently, I missed to create a pull request back last month.

### DIFF
--- a/new/LIT7273PPEntering.xml
+++ b/new/LIT7273PPEntering.xml
@@ -1,0 +1,79 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7273PPEntering" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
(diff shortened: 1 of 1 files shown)


karljonaskarlsson (approved):

new/LIT7273PPEntering.xml
```diff
@@ -0,0 +1,79 @@
...
+                    <ab>አግብኦ፡ ወመልሶ፡ ለገብረ፡ እገሌ፡ ውስተ፡ ቤተ፡ እገሌ፡</ab>
+                </div>
+            </div>
+            <div type="translation" xml:lang="en">
+                <note>Translation by Carsten Hoffmann with the support by Jonas Karlsson.</note>
+                <div type="textpart" subtype="incipit">
+                    <ab>May they burn (?) ʾAlʾazina (?)</ab>
+                </div>
+                <div type="textpart" subtype="explicit">
+                    <ab>... making to return and flashing the servant so and so in the house of so and so.</ab>
```
thea-m, line 74: I don't understand what is meant by "flashing". I would probably translate "return" (maybe transitively under Amharic influence?)

thea-m (approved):

new/LIT7273PPEntering.xml
CarstenHoffmannMarburg: I do not know, what made me translate "flashing". I may have been confused with another record.

I want to avoid using the word "return" twice. What do you think about the following translation:

"... making (him) to go back and to return, (to the benefit of) the servant so and so in the house of so and so."

If it is too complicated and we are not sure about the meaning, I may also delete the translation of the explicit.

@MagdaKrzyz , what is your opinion about this phrase and translation?

MagdaKrzyz: I would say: to bring him back and return him to the servant so and so to the house of so and so

PR #1518, opened 2025-03-25
Update edition in LIT7321InfantsSus

CarstenHoffmannMarburg:
According to our discussions in https://github.com/BetaMasaheft/Works/pull/1506#issuecomment-2751102726.

### DIFF
--- a/new/LIT7321InfantsSus.xml
+++ b/new/LIT7321InfantsSus.xml
@@ -55,6 +55,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         </profileDesc>
         <revisionDesc>
             <change who="CH" when="2025-03-19">Created entity</change>
+            <change when="2025-03-24" who="CH">Updated edition according to TEI</change>
         </revisionDesc>
     </teiHeader>
     <text>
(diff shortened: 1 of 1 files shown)


new/LIT7321InfantsSus.xml
```diff
@@ -67,19 +68,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </listRelation>
             </div>
             <div type="edition">
-                <note>The following incipit is taken from <ref target="#E"/> with some reading from
-                    <ref target="#B"/> according to the catalogue description in 
+                <note>The following incipit is taken from <ref type="mss" corresp="#E"/> with some reading from
```
eu-genia, line 71: You cant use corresp for internal references

eu-genia (changes requested):

new/LIT7321InfantsSus.xml
CarstenHoffmannMarburg: If I cannot use `<ref target="#E"/>` and I cannot use `<ref type="mss" corresp="#E"/>`, what can I use? It would be very helpful, if you could give me a hint rather than pointing to the guidelines in general.

eu-genia: of course you can use `target `for references to anchors with `# `and corresp for external, just you cannot expect that the shelfmark will be expanded if you point to an internal anchor, the maximum one should expect is the id (sigla) printed (if it is not visible then the code might need to be adjusted)

eu-genia (approved):

PR #1545, opened 2025-04-17
Dr 25 04 miracles

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2942

### DIFF
--- a/new/LIT7347MGMTazkar.xml
+++ b/new/LIT7347MGMTazkar.xml
@@ -1,0 +1,60 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7347MGMTazkar" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">Miracle of Gabra Masqal: Miracle worked during the tazkār organized on the death day of the saint</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DR"/>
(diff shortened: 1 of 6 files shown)


new/LIT7347MGMTazkar.xml
```diff
@@ -0,0 +1,83 @@
...
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>First miracle in <ref type="mss" corresp="ANLcr11"/>, paraphrased according to
+                    the descriptions of Strelcyn and Raineri:
+                   An obscure miraculous event at
+                   the <foreign xml:lang="gez">tazkār</foreign> organized on the death day of the saint.</p>
```
nafisa-valieva, line 37: here too, if it is what they wrote and you want to keep it, could you just quote them?

new/LIT7347MGMTazkar.xml
```diff
@@ -0,0 +1,83 @@
...
+                        <citedRange unit="page">34</citedRange>
+                    </bibl>
+                </listBibl>
+            </div>
+            <div type="edition">
+                <div type="textpart" subtype="incipit" xml:lang="gez">
+                    <label>Incipit</label>
+                    <note>The following incipit is taken from <ref type="mss" corresp="ANLcr11"/>.</note>
+                    <ab>፩፡ ተአምሪሁ፡ ለአቡነ፡ ገብረ፡ መስቀል። ወእምዝ፡ በ፵፡ ዕለት፡ እምዘአዕረፈ፡ ተጋብኡ፡ ብዙኃን፡ ሰብእ፡ እምለፌ፡
+                        ወለፌ፡ ከመ፡ ይግበሩ፡ ተዝካሮ፡ ወይንሥኡ፡ በረከቶ፡ ወአኃውኒ፡ መነኮሳት፡ እንዘ፡ ያስትዳልው፡ ኅብስተ፡ ወጽዋዓ፡ ለተዝካር፡ ከመ፡
```
nafisa-valieva, line 75: ያስትዳልዉ፡?

new/LIT7347MGMTazkar.xml
thea-m: That would also be possible! in this case, I think it keeps the record more concise to paraphrase, and the added information from quoting two different, but very similar descriptions would not be worth the added quantity of text

MagdaKrzyz (approved):
Thanks!

PR #1578, opened 2025-04-30
Added change notes to qene records

CarstenHoffmannMarburg:
I added change notes to all qene records, that I have created so far, as it was requested in https://github.com/BetaMasaheft/Documentation/issues/2700 in order not to indicate them as "stubs".

I also made some minor updates (deleted source desc, attestation in div, added keyword "Christian Literature" and changed on abstract ("indicated as" to avoid "indicated in incipit" as requested by Denis in another pull request).

### DIFF
--- a/7001-8000/LIT7047Qene.xml
+++ b/7001-8000/LIT7047Qene.xml
@@ -23,9 +23,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </availability>
             </publicationStmt>
             <sourceDesc>
-                <listWit>
-                    <witness corresp="BNFabb145"></witness>
-                </listWit> </sourceDesc>
+                <p/>
+            </sourceDesc>
         </fileDesc>
         <encodingDesc>
             <p>A digital born TEI file</p>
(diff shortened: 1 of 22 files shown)


7001-8000/LIT7047Qene.xml
```diff
@@ -23,9 +23,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </availability>
             </publicationStmt>
             <sourceDesc>
-                <listWit>
-                    <witness corresp="BNFabb145"></witness>
-                </listWit> </sourceDesc>
```
eu-genia, line 28: why deleted here and elsewhere? both here and relation? I would leave the witness

karljonaskarlsson (approved):

eu-genia (approved):

7001-8000/LIT7047Qene.xml
CarstenHoffmannMarburg: I thougt it was decided not to mention the manuscript in listWit as far as they are mentioned in a note in the div. Shall I make these changes undone? Should we not better have uniformity for all the qene records?

eu-genia: Manuscripts used in the `edition `section of the work records MUST be listed in `witnesses`, we had discussed it many many times

eu-genia: What I do not recommend is doubling the information in relation, this is true.

See also here https://github.com/BetaMasaheft/Documentation/issues/2699 and my observation on search results.

PR #1580, opened 2025-05-05
refactored and fixed

eu-genia:
replacing [#1495](https://github.com/BetaMasaheft/Works/pull/1495) no. 2


eu-genia:
Wait do not merge
I may have overwritten something, will have to redo

PR #1581, opened 2025-05-05
refactored and fixed

eu-genia:
replacing #1495 no. 3


eu-genia:
Wait do not merge
I may have overwritten something, will have to redo

PR #1594, opened 2025-05-13
Created LIT7426HomilyBenj and LIT7427VitaSua

CarstenHoffmannMarburg:
I created two new records for the text in PetermannIINachtrag24 and its incorporated test of "Narratio de vita sua" as proposed in https://github.com/BetaMasaheft/Documentation/issues/2944.

I updated LIT5812OntheC as proposed by Dorothea in https://github.com/BetaMasaheft/Documentation/issues/2678.

For indicating author and Clavis ID, I followed the example of LIT2122OntheF, even though, personally, I do not like to indicate CPG6633 both in the title and in source desc, because I think, the first indication in title is redundant.

### DIFF
--- a/new/LIT7426HomilyBenj.xml
+++ b/new/LIT7426HomilyBenj.xml
@@ -1,0 +1,75 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7426HomilyBenj" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:id="t1">Homily of Benjamin of Alexandria for Easter Tuesday</title>
+                <author ref="PRS2659Benjamin"/>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Homily of <persName ref="PRS2659Benjamin"/> on the Resurrection for Easter Tuesday, in which is embedded the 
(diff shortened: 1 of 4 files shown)


new/LIT7426HomilyBenj.xml
```diff
@@ -0,0 +1,75 @@
...
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Homily of <persName ref="PRS2659Benjamin"/> on the Resurrection for Easter Tuesday, in which is embedded the 
+                    <ref type="work" corresp="LIT7427VitaSua"/> of <persName ref="PRS7973PseudoD"/>. Copies of this homily offer the 
+                    earliest Geʿez evidence for this apocryphon by far, but a definitive answer to the important question of whether the Narratio de vita
```
nafisa-valieva, line 36: 'by far' does not sound correct here for me

new/LIT7426HomilyBenj.xml
```diff
@@ -0,0 +1,75 @@
...
+                    sua was interpolated into the Ethiopic version of the homily soon after its translation or belonged to the original translated text and
```
nafisa-valieva, line 37: I am not sure this formulation is clear. At least it is not very clear to me. Your question is whether the text was translated as part of a homily or alone, yes?

new/LIT7426HomilyBenj.xml
```diff
@@ -0,0 +1,75 @@
...
+                </keywords>
+            </textClass>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="grc">Greek</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2025-05-13">Created entity</change>
+            <change when="2025-05-18" who="CH">Corrections after reviews by Dorothea Reule and Nafisa Valieva</change>
```
nafisa-valieva, line 56: also for the change element here and elsewhere: I might have forgotten, then correct me, but have we reviewed these records before? My memory has no traces of it!

nafisa-valieva (approved):
Thank you!

new/LIT7426HomilyBenj.xml
CarstenHoffmannMarburg: LIT7427VitaSua is part of LIT7426HomilyBenj. The question is, whether it entered the text before or after its translation.

I took the sentence from Ted Erho. I can delete, if it is not clear enough.

new/LIT7426HomilyBenj.xml
CarstenHoffmannMarburg: I have inserted the second change note in order not to forget it in the end, because I do not want to end it up as draft or stub. I will alter this method, if it is confusing to you.

nafisa-valieva: > I have inserted the second change note in order not to forget it in the end, because I do not want to end it up as draft or stub. I will alter this method, if it is confusing to you.

it is ok, no worries! I was just curious! Just remember to modify it if reviewers do not have comments!!!

new/LIT7426HomilyBenj.xml
nafisa-valieva: > LIT7427VitaSua is part of LIT7426HomilyBenj. The question is, whether it entered the text before or after its translation.
> 
> I took the sentence from Ted Erho. I can delete, if it is not clear enough.

Maybe mark it as quotation from his article! I think it would be better like this!

new/LIT7427VitaSua.xml
```diff
@@ -0,0 +1,84 @@
...
+                <language ident="la">Latin</language>
+                <language ident="grc">Greek</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2025-05-13">Created entity</change>
+            <change when="2025-05-18" who="CH">Corrections after reviews by Dorothea Reule and Nafisa Valieva</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
```
thea-m, line 63: I think it would be good to include the Geez incipit in this file (and maybe also the others, if possible, for exmaple from bm:Muthreich2022Dionysius, p. 32-33). Especially in combination with the latin title, I think this file might be currently a bit difficult to parse for anyone not familiar with the literature

thea-m (approved):

PR #1626, opened 2025-05-28
updated the record

nafisa-valieva:
bm:Heide2003EAEGadlaA is very problematic: it says that Conti Rossini's edition is based on one manuscript without saying on which; the sources and literature seem to be mixed; more recent edition, `  <bibl>
                        <ptr target="bm:Aescoly1951Recueil"/>
                        <citedRange unit="page">51-77</citedRange>
                    </bibl>` is not mentioned at all. I wonder whether we have a place to say it and whether this critical note is needed at all.


thea-m (approved):
Very nice improvement! In principle, a  `<note>` inside `<bibl>` is possible, though I don't know if it will be visualized. My feeling would be to refrain from this, because such notes are not normally added to all records. Another possibility to alert future colleagues in a less public way would be to write a comment inside the file.

PR #1640, opened 2025-06-03
Create LIT7496Sawwer.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2998.

The title and the incipit was not very informative to me. It seems, that there is a lot of astrological or magical terminology, that I cannot understand or translate. For this reason, I preferred to keep Kalkalos, Zāru and Ṭimos as proper names. 

Between these enigmatic terms, there are some words (especially የሚባል), that remind to Amharic, and some, which are Gəʾəz. I identified it therefore as a Gəʾəz-text, which is "partly Amharic". 

I do not know, what the term sawwər (= conceal/hide, Imperative) stands for. It might be the case, that we are informed about conjunctions, whereas one planet will 'conceal' another or any other magic meaning of the word.

### DIFF
--- a/new/LIT7496Sawwer.xml
+++ b/new/LIT7496Sawwer.xml
@@ -32,7 +32,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>An astrological treatise in Gǝʾǝz and partly in Amharic, indicating the diseases corresponding to the twelve signs of the zodiac, 
+                <p>Astrological treatise in Gǝʾǝz and partly in Amharic, indicating the diseases corresponding to the twelve signs of the zodiac, 
                     along with their respective remedies.</p>
             </abstract>
             <textClass>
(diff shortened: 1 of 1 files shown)


new/LIT7496Sawwer.xml
```diff
@@ -0,0 +1,66 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7496Sawwer" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">፪ ሰውር፡ ወስመ፡ ደዌሁ፡ ከልከሎስ፡ ዛሩ፡ ጢሞስ፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">2 sawwǝr wa-sǝma dawehu Kalkalos Zāru Ṭimos</title>
+                <title xml:lang="en" corresp="#t1">Two, conceal! Name of his disease, Kalkalos, Zāru, Ṭimos</title>
```
MagdaKrzyz, line 10: And the name of his... 
I have no idea what this ፪ ሰውር፡ means

MagdaKrzyz: Good

PR #1662, opened 2025-06-12
Create records for Maṣḥafa Madḫanit

CarstenHoffmannMarburg:
I created two new work records for distinct versions of the Maṣḥafa Madḫanit and updated the existing one (LIT1942Mashaf) according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3017.

I added the keyword Divination to LIT1171Awdana.

PR #1663, opened 2025-06-12
Create LIT7531Eye.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3025 with similar prescriptions as in https://github.com/BetaMasaheft/Works/pull/1662.

### DIFF
--- a/new/LIT7531Eye.xml
+++ b/new/LIT7531Eye.xml
@@ -1,0 +1,80 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7531Eye" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="en" xml:id="t1">Medical prescriptions in Amharic</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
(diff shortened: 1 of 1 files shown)


new/LIT7531Eye.xml
```diff
@@ -0,0 +1,80 @@
...
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
+                    <term key="Magic"/>
+                    <term key="Medicine"/>
+                </keywords>
+            </textClass>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="am">Amharic</language>
```
MagdaKrzyz, line 44: and geez

new/LIT7531Eye.xml
```diff
@@ -0,0 +1,80 @@
...
+            <div type="edition" xml:lang="am">
+                <note>The division in textparts and the incipits are taken from <ref type="mss" corresp="BLorient12034"/> according to the 
+                    description in <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="item">98,9</citedRange>
+                        <citedRange unit="page">145</citedRange></bibl>.</note>
+                <div type="textpart" subtype="chapter" xml:id="Haem">
+                    <label>Prescription against vaginal haemorrhage</label>
+                    <ab>የደም፡ <gap reason="ellipsis"/> ለዘይውኅዝ፡ ትክቶሃ፡</ab>
+                </div>
+                <div type="textpart" subtype="chapter" xml:id="Foetus">
+                    <label>Preescription for protecting the foetus</label>
```
thea-m, line 63: Prescription

new/LIT7531Eye.xml
```diff
@@ -0,0 +1,80 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Prescriptions taken from a collection of medical prescriptions (maṣḥafa madḫanit).</p>
```
thea-m, line 33: Does this record consistute a textpart of one of the created maṣḥafa madḫanit records? If so, it would make sense to include a relation element in the record. I also think that if these prescriptions circulate under the label "maṣḥafa madḫanit", it would be good to include this in fidal somewhere in the record for searchability reasons.

thea-m (changes requested):
Thank you, very nice! It seems to be a complex situation, these records will help a lot

new/LIT7531Eye.xml
CarstenHoffmannMarburg: I think, Strelcyn sees all medical prescriptions in collections as maṣḥafa madḫanit and he recalls them either under this name or as "traité thérapeutique". The very few prescriptions listed here are not necessarily from the same source as those recorded in https://github.com/BetaMasaheft/Works/pull/1662, unless that cannot be excluded. I suggest to create relations (dc:relation) to mark those as similar, which concern the same body parts (e.g. the eye) as I have indicated in a note, already.

CarstenHoffmannMarburg:
> Thank you, very nice! It seems to be a complex situation, these records will help a lot

Yes, but it seems to me to be just the tip of the iceberg.

MagdaKrzyz (approved):

thea-m (approved):

PR #1668, opened 2025-06-16
Create LIT7534PPThieves.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3024.

### DIFF
--- a/new/LIT7534PPThieves.xml
+++ b/new/LIT7534PPThieves.xml
@@ -6,8 +6,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <fileDesc>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">ጸሎት፡ በእንተ፡ ሠራቂ፡ ወቀመኛ፡</title>
-                <title xml:lang="gez" type="normalized" corresp="#t1">Ṣalot baʾǝnta śarāqi wa-Qamaññā</title>
-                <title xml:lang="en" corresp="#t1">Prayer against thieves and Qamaññā</title>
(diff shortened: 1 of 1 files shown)


new/LIT7534PPThieves.xml
```diff
@@ -0,0 +1,73 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7534PPThieves" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ጸሎት፡ በእንተ፡ ሠራቂ፡ ወቀመኛ፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Ṣalot baʾǝnta śarāqi wa-Qamaññā</title>
```
MagdaKrzyz, line 9: If you know what Qamaññā is why not translating as "sorcerers"? I understand that this is not a proper name so it should not be capitalized. The explanation that you give below is good and enough.

MagdaKrzyz (approved):

new/LIT7534PPThieves.xml
CarstenHoffmannMarburg: OK.

PR #1669, opened 2025-06-16
Create LIT7535Kobar.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3026.


MagdaKrzyz (approved):

thea-m (changes requested):
Great, thank you! Could you also include the variant "kobor" either in the title or abstract in fidal and transliteration, so that users can find this record when searching for it?

thea-m (approved):

PR #1679, opened 2025-06-23
Create LIT7541PPAhya.xml

CarstenHoffmannMarburg:
According to my suggestion in https://github.com/BetaMasaheft/Documentation/issues/3037#issuecomment-2990764861.

### DIFF
--- a/new/LIT7541PPAhya.xml
+++ b/new/LIT7541PPAhya.xml
@@ -1,0 +1,84 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7541PPAhya" xml:lang="en" type="work">
+    <teiHeader>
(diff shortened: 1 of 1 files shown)


new/LIT7541PPAhya.xml
```diff
@@ -0,0 +1,84 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7541PPAhya" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">በል፡ አኽያ፡ ሸራኽያ፡</title>
+                <title xml:lang="gez" type="normalized" corresp="#t1">Bal ʾaḵyā šarāḵyā ...</title>
```
MagdaKrzyz, line 9: አኽያ፡ could be pronounced the same as አህያ "donkey", thus  ʾaḵǝyyā, and then also  šarāḵǝyyā
What do you think?

MagdaKrzyz (approved):

new/LIT7541PPAhya.xml
CarstenHoffmannMarburg: Yes, I will change the transcription and also the translation and the xml:lang, because the title is Amharic, then.

PR #1680, opened 2025-06-23
Create LIT7542MasAmero.xml

CarstenHoffmannMarburg:
Cf. https://github.com/BetaMasaheft/Documentation/issues/3040#issuecomment-2995821573.

### DIFF
--- a/new/LIT7542MasAmero.xml
+++ b/new/LIT7542MasAmero.xml
@@ -32,11 +32,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>Treatise of divination divided into twelve chapters.</p>
+                <p>Treatise of divination divided into twelve chapters as indicated in the incipit.</p>
             </abstract>
             <textClass>
                 <keywords>
                     <term key="Magic"/>
+                    <term key="Divination"/>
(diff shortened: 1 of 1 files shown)


new/LIT7542MasAmero.xml
```diff
@@ -0,0 +1,67 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Treatise of divination divided into twelve chapters.</p>
```
nafisa-valieva, line 35: are they always divided into twelve chapters? if not, may be as in the abstract which ms you used as a source.

new/LIT7542MasAmero.xml
```diff
@@ -0,0 +1,67 @@
...
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="Magic"/>
```
nafisa-valieva, line 39: also ChristianLiterature, no?

nafisa-valieva (approved):
Nice!

new/LIT7542MasAmero.xml
CarstenHoffmannMarburg: This information is given in the incipit below. I will clarify in the abstract.

PR #1683, opened 2025-06-25
Medical prescriptions in Amharic

CarstenHoffmannMarburg:
I created four records for medical prescriptions against various disease and also some demons. Mostly in Amharic. According to my suggestion in https://github.com/BetaMasaheft/Documentation/issues/3036.

### DIFF
--- a/new/LIT7543Prescripti.xml
+++ b/new/LIT7543Prescripti.xml
@@ -53,7 +53,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         </profileDesc>
         <revisionDesc>
             <change who="CH" when="2025-06-23">Created record</change>
-            <change when="2025-06-25" who="CH">Corrections after review by Magdalena Krzyżanowska</change>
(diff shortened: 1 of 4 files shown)


CarstenHoffmannMarburg:
Thank you for your informative comments! If all your requests are met, can you approve?

nafisa-valieva:
Carsten, I think it would be nice to verify these lables with native speakers. If you refuse it, I will simply not review it.

CarstenHoffmannMarburg:
> Carsten, I think it would be nice to verify these lables with native speakers. If you refuse it, I will simply not review it.

I think, this case is not worth to involve Solomon, Mersha or Guesh. Medical prescriptions are not part of their research interests and it will take too much or their time and will have no profit to them. If none of us feels competent enough to review the labels, I would rather prefer to delete the labels or the textparts overall and keep the text without my own comment in the manuscript record. As I said above, if we do not know, we can always refer to Stefan Strelcyn, who already gave his opinion in a printed catalogue.

nafisa-valieva (approved):

PR #1716, opened 2025-07-28
Update Tarika Walda Amid records

CarstenHoffmannMarburg:
I updated the listWit of the two recensions moving ms O from α to β after having a closer look to that manuscript.


nafisa-valieva (approved):

CarstenHoffmannMarburg:
I found that the text in ESdd031 is an attestation of LIT7447AbrTarWA.

I updated the relations LIT6935TarWaAmidAlpha and LIT6936TarWaAmidBeta to disambiguate them from LIT7447AbrTarWA.

PR #1749, opened 2025-08-20
Create LIT7592MedicPre.xml

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3095.

To be checked with consultation by a native speaker as suggested by Magda.

PR #1752, opened 2025-08-20
Update abstract (asent) of LIT7592MedicPre

CarstenHoffmannMarburg:
According to your request in https://github.com/BetaMasaheft/Works/pull/1749#discussion_r2288057822.

### DIFF
--- a/new/LIT7592MedicPre.xml
+++ b/new/LIT7592MedicPre.xml
@@ -30,7 +30,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>Medical prescriptions in Amharic for a woman who cannot or does not want to have orgasm 
+                <p>Medical prescriptions in Amharic for a woman who is avoiding sexual relations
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
I have changed the record according to your suggestion. Would you like to approve it?

new/LIT7592MedicPre.xml
```diff
@@ -30,7 +30,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <profileDesc>
             <creation/>
             <abstract>
-                <p>Medical prescriptions in Amharic For a woman who became skinny or for a woman who cannot or does not want to have orgasm 
+                <p>Medical prescriptions in Amharic for a woman who cannot or does not want to have orgasm
```
MagdaKrzyz, line 33: Actually, I suggested that it means simply "for a woman who avoids sexual relations".

MagdaKrzyz (changes requested):

new/LIT7592MedicPre.xml
CarstenHoffmannMarburg: Perhaps we have downplayed the Amharic phrase in this translation. However, as I am not an expert in Amharic, I will trust your expertise and change it according to your preferred phrase.

PR #1759, opened 2025-08-22
Create LIT7600HassMamer

CarstenHoffmannMarburg:
I created LIT7600HassMamer according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2999#issuecomment-3195847809.

I added the keyword divination to previously created records.

### DIFF
--- a/new/LIT7600HassMamer.xml
+++ b/new/LIT7600HassMamer.xml
@@ -8,7 +8,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <title xml:lang="gez" xml:id="t1">ሐሳበ፡ ማእምር፡</title>
                 <title xml:lang="gez" corresp="#t1" type="normalized">Ḥassāba māʾmǝr (general record)</title>
                 <title xml:lang="en" corresp="#t1">Calculation of the learned man</title>
-                <title xml:lang="en" corresp="t2">Divination regarding learned men</title>
+                <title xml:lang="en" corresp="t2">Divination regarding the learned man</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="CH"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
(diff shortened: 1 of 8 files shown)


new/LIT7600HassMamer.xml
```diff
@@ -0,0 +1,61 @@
...
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7600HassMamer" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">ሐሳበ፡ ማእምር፡</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">Ḥassāba māʾmǝr (general record)</title>
+                <title xml:lang="en" corresp="#t1">Calculation of the learned man</title>
+                <title xml:lang="en" corresp="t2">Divination regarding learned men</title>
```
MagdaKrzyz, line 11: Why do you have here a plural "men" and above singular? I think "regarding the learned man" is better. Then we have a class and we do not have to think whether there is one guy meant or many.

new/LIT7600HassMamer.xml
```diff
@@ -0,0 +1,61 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p>Divinatory text (ḥassāb) concerning learned men, to be used when no further information is retrievable.</p>
```
MagdaKrzyz, line 36: the same as above

MagdaKrzyz (approved):

PR #1764, opened 2025-08-25
Create LIT7605PPIllnesses.xml

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/3104

### DIFF
--- a/new/LIT7605PPIllnesses.xml
+++ b/new/LIT7605PPIllnesses.xml
@@ -1,0 +1,70 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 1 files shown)


new/LIT7605PPIllnesses.xml
```diff
@@ -0,0 +1,70 @@
...
+                    <term key="Prayers"/>
+                </keywords>
+            </textClass>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="gez">Gǝʿǝz</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="DR" when="2025-08-25">Created stub</change>
```
CarstenHoffmannMarburg, line 51: I would not call this a stub, but rather a full record.

new/LIT7605PPIllnesses.xml
```diff
@@ -0,0 +1,70 @@
...
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
+                    <term key="Magic"/>
+                    <term key="Prayers"/>
```
CarstenHoffmannMarburg, line 42: I think, we can count qāl as similar to ʾasmāt and use the keyword "Asmat" here as well, even though the asmat in this prayer are hidden in the gap.

PR #1770, opened 2025-08-25
Lit1292 dersan corr mk 23.08.25

MarcinKrawczuk:
This is a basic record inherited from the EAe (I've checked all the instances where it is used in the EAe and I'm 98% sure about the identification), so I've brushed it up a bit, because I think this mss. is very important.


CarstenHoffmannMarburg:
> This is a basic record inherited from the EAe (I've checked all the instances where it is used in the EAe and I'm 98% sure about the identification), so I've brushed it up a bit, because I think this mss. is very important.

Wonderful! 

Note: I believe that the remaining 2% likelihood of error is adequately reflected in the abstract.

PR #1782, opened 2025-08-28
Dr 25 08 libanos

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/3078


MagdaKrzyz:
Actually, I liked your phrase "which is the most frequently/widely attested version of this work."

PR #1806, opened 2025-09-04
Two new malke' records (see #3118)

smaugustine:
Two new malkǝʾ records (per https://github.com/BetaMasaheft/Documentation/issues/3118).


CarstenHoffmannMarburg:
Very nice!

Thank you!

PR #1813, opened 2025-09-09
updates to Malk4Ensesa

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/3119


smaugustine:
If you like, I have also the second stanza (which helps disambiguate) and the final stanza (as explicit): https://malkeagubae.com/api/xml/works/2935.xml

PR #1814, opened 2025-09-10
Confessions of faith

MassimoVilla:
One new record for a Confession of faith, one general record and few updates about relations in existing records

https://github.com/BetaMasaheft/Documentation/issues/3131


new/LIT7646Creed.xml
```diff
@@ -0,0 +1,65 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="MV" when="2025-09-10">Created record</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <div type="bibliography">
+                <listRelation>
+                    <relation name="saws:isVersionOf" active="LIT7615GadLib1" passive="LIT1473GadlaL"/>
```
eu-genia, line 58: I guess the active should be LIT7646Creed

eu-genia (changes requested):

new/LIT7646Creed.xml
MassimoVilla: Those relations shouldn't be there at all, sorry.
Thank you

eu-genia (approved):

PR #1836, opened 2025-09-24
Update LIT1830Mahbar.xml

CarstenHoffmannMarburg:
I updated LIT1830Mahbar with information from EAe and the catalogue description by Strelcyn for BLorient13309, which includes a lengthy incipit. 

In this manuscript, t1 is explicitly mentioned, while t2 is absent. In the EAe article by Samuel Yalew, both titles are given without indication of priority. I have prioritised t1 over t2, assuming that the former is the original title and the latter a secondary one.


CarstenHoffmannMarburg:
@MagdaKrzyz , can you review this pull request of a deletion of LIT1830Mahbar and an update of LIT2001Mazmur?

PR #1839, opened 2025-10-01
Magda dd 031 wegat

MagdaKrzyz:


new/LIT7711SalamSharpPain.xml
```diff
@@ -0,0 +1,69 @@
...
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
```
CarstenHoffmannMarburg, line 35: I am still wondering, whether this is infact a magic prayer or rather a poem, which welcomes piercing for whatever reason. You could perhaps add an abstract to express this incertainty, e.g.:
<p>Kind of a salām-poem, which was possibly used for magic purposes.</p>

CarstenHoffmannMarburg (approved):

new/LIT7711SalamSharpPain.xml
MagdaKrzyz: Could be but it is written together with two other magic prayers.

MagdaKrzyz: I have removed "magic"

PR #1841, opened 2025-10-03
Magda dd 031 works corrections

MagdaKrzyz:
I have made some changes but the main problem is that the previous merge did not function.


CarstenHoffmannMarburg:
> I have made some changes but the main problem is that the previous merge did not function.

Which merge did not work out? I can't see any closed branches or failed merges.

MagdaKrzyz:
You cannot see it but once you check the app you can see that the content is simply not there.

eu-genia:
You cannot do any merges when the app is offline, they cannot get there, so if this one was merged with server offline it won't be there.

MagdaKrzyz:
The server was not offline and this was my second attempt. The first one was on a different day.

PR #1843, opened 2025-10-06
Update

MagdaKrzyz:
let me try once again. Could you just approve the merge?


eu-genia:
Please do notify immediately if not all get through as it means i must fix something. No need to wait, waiting is only if titles or bibliography are not printed, not for normal changes

PR #1988, opened 2026-04-20
Delete doublet LIT7809MaggMir, update LIT7194MGAbraset

CarstenHoffmannMarburg:
According to my suggestion in https://github.com/BetaMasaheft/Documentation/issues/3301.

PR #1990, opened 2026-04-20
Added xml:id for editions in LIT7194MGAbraset

CarstenHoffmannMarburg:
I created xml:id for editions as @eu-genia has requested  in https://github.com/BetaMasaheft/Works/pull/1988/changes/a703cf8c2027ce99d8760c325fec7c31452508e4#r3109457638.

I tried to avoid the id ANLcr3, because I think that this would possibly interfere with the ID of the manuscript.


MagdaKrzyz (approved):

eu-genia (approved):
it is OK (correct to not use existing IDs, would have been ok also with "ed1" "ed2" or anything at all.

would have been even better to do that in the same PR where the comment was in https://github.com/BetaMasaheft/Persons/pull/1351, before merging.

e-schwarz (approved):

PR #1992, opened 2026-04-23
Magda dd 042 works

MagdaKrzyz:
I have moved witnesses from ESdd042. 
There seems to be a problem with the visualization of the textparts. Only some of them appear in the app.

### DIFF
--- a/2001-3000/LIT2598Yosipp.xml
+++ b/2001-3000/LIT2598Yosipp.xml
@@ -36,6 +36,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <p>Work of the literatures of Ethiopia and Eritrea</p>
             <p>A TEI file converted from the ETHIO-authority google spreadsheet</p>
          </sourceDesc>
+        <sourceDesc>
+        <listWit>
+          <witness corresp="EMML4773">(16th/17th cent.)</witness>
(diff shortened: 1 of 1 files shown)


2001-3000/LIT2598Yosipp.xml
```diff
@@ -36,6 +36,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <p>Work of the literatures of Ethiopia and Eritrea</p>
             <p>A TEI file converted from the ETHIO-authority google spreadsheet</p>
          </sourceDesc>
+        <sourceDesc>
+        <listWit>
+          <witness corresp="EMML4773">Dated to 16th/17th century</witness>
```
eu-genia: if you write the text here you will only have the text printed and NOT THE SHELFMARK.

the rule is always the same: either computed visualization, if the element is empty, or the content of the element. if you want the date to be visible then you have to repeat the shelfmark.

eu-genia (changes requested):
please refresh the file (fetch-pull) in github desktop before editing as i had modified the edition.


the visualization is now fixed, thank you.

2001-3000/LIT2598Yosipp.xml
MagdaKrzyz: right

eu-genia (approved):

PR #1996, opened 2026-04-29
Lit6402 mj infancy corr mk 29.04.26

MarcinKrawczuk:
I thought that maybe now that the miracles are in the database we could link them to the overarching record. If it's a bad idea, I'll delete the changes.


CarstenHoffmannMarburg:
> I thought that maybe now that the miracles are in the database we could link them to the overarching record. If it's a bad idea, I'll delete the changes.

It is a valuable addition to the record and even a necessary one.

CarstenHoffmannMarburg (approved):

PR #2007, opened 2026-05-11
Magda aragawi menfesawi

MagdaKrzyz:


CarstenHoffmannMarburg (approved):

eu-genia:
It should be xml:id=t3 and not corresp=t3, now the title is not visible because of the encoding error

MagdaKrzyz:
Yes, I have noticed. I will correct the file soon.

PR #2008, opened 2026-05-13
Magda makm 014 works

MagdaKrzyz:


eu-genia (approved):

MagdaKrzyz:
@eu-genia, could you look at the visualization of LIT5625TergSenaFet?
The second relation is missing:
`<relation name="ecrm:P129_is_about" active="LIT1546Genesi"><desc>Interpretation of <ref type="work" corresp="LIT1546Genesi#Gen1">the Book of Genesis, Chapter 1</ref>.</desc></relation>`

And the second incipit is also not there. 
I cannot find any mistakes in the encoding.

eu-genia:
the relation should have active 5625 and passive 1546.
both incipits are expected to be shown under text tab, the main page only prints one first incipit in accordance with the requirement

eu-genia:
text is not shown as the edition is inside of div bibliography

MagdaKrzyz:
Thank you! I will correct the relation.
As for 
> text is not shown as the edition is inside of div bibliography
explain please: the first is also inside of div. I thought it should be displayed in Text.

eu-genia:
the snippet code looks for `any `div marked as incipit to display it.

the text only displays div type edition, not if it is hidden somewhere.

once you have fixed the encoding the text will appear.

eu-genia:
(but the first page snippet will still only show ONE incipit)

PR #2024, opened 2026-06-12
Create LIT7865PrSalomon.xml

CarstenHoffmannMarburg:
According to my suggestion in https://github.com/BetaMasaheft/Documentation/issues/3334.

### DIFF
--- a/new/LIT7865PrSalomon.xml
+++ b/new/LIT7865PrSalomon.xml
@@ -7,7 +7,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <titleStmt>
                 <title xml:lang="gez" xml:id="t1">አስማተ፡ ሰሎሞን፡ ጽድርያስ፡ አያስ፡</title>
                 <title xml:lang="gez" corresp="#t1" type="normalized">ʾAsmāta Salomon ṣǝdǝryān ʾayās</title>
-                <title xml:lang="en" corresp="#t2">Names of Solomon: ṣǝdǝryān ʾayās</title>
+                <title xml:lang="en" corresp="#t1">Names of Solomon: ṣǝdǝryān ʾayās</title>
                 <title xml:lang="en" xml:id="t2">Prayer against blacksmiths and magicians by the names of Solomon</title>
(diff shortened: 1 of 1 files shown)


new/LIT7865PrSalomon.xml
```diff
@@ -0,0 +1,78 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="LIT7865PrSalomon" xml:lang="en" type="work">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title xml:lang="gez" xml:id="t1">አስማተ፡ ሰሎሞን፡ ጽድርያስ፡ አያስ፡</title>
+                <title xml:lang="gez" corresp="#t1" type="normalized">ʾAsmāta Salomon ṣǝdǝryān ʾayās</title>
+                <title xml:lang="en" corresp="#t1">Names of Solomon: ṣǝdǝryān ʾayās</title>
```
MagdaKrzyz: If these are his names, perhaps it would be better to capitalize them?

MagdaKrzyz (approved):

new/LIT7865PrSalomon.xml
CarstenHoffmannMarburg: You are right. This may be confusing to many users. However, these are probably not personal names, but magic words. We normally have a difference between magic words and personal names as for example in LIT7445PPQahashun. Is it better to translate the word ʾAsmāta with 'Secret words' or 'Magic words'?

MagdaKrzyz: Are they secret if they are pronounced?

CarstenHoffmannMarburg: OK, I will use Asmat instead.

PR #2046, opened 2026-06-30
Update LIT3186Meeraf.xml

CarstenHoffmannMarburg:
In each relation in relation desc, the active can only be filled by the ID of the record and the passive is filled by the ID of what it shall refer to. As far as I know, we use saws:contains, if a text does normally or may possibly contain another text.


MagdaKrzyz (approved):
Right!

PR #2063, opened 2026-07-16
Update LIT3063RepCh343.xml

e-schwarz:


e-schwarz:
I noticed that in some witnesses this hymn starts a bit differently, so I added the second incipit to make the search easier

CarstenHoffmannMarburg (approved):

